PR #11, opened 2024-05-30
ci: GitHub Action to detect typos with codespell

cclauss:
https://github.com/codespell-project/actions-codespell

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -6,6 +6,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: codespell-project/actions-codespell@master
+      - uses: codespell-project/actions-codespell@v2
         with:
           ignore_words_list: crate,raison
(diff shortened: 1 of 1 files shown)


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,11 @@
+# https://github.com/codespell-project/actions-codespell
+name: codespell
+on: [pull_request, push]
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: codespell-project/actions-codespell@master
```
AugustinMauroy, line 9: we should use specific version.

AugustinMauroy:
i'm +1 for code spell

.github/workflows/codespell.yml
cclauss:
```suggestion
      - uses: codespell-project/actions-codespell@v2
```

AugustinMauroy (approved):
LGMT !

PR #126, opened 2024-10-15
chore(dependabot): increase interval

AugustinMauroy:
## Description

As proposed on slack, we should reduce the number of dep pr.

### Check List

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
 # We define the interval for the updates to be monthly
-# because we doesn't want to have too many updates
+# because we don't want to have too many updates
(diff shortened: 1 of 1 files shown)


flakey5 (approved):

.github/dependabot.yml
```diff
@@ -1,17 +1,21 @@
+# We define the interval for the updates to be monthly
+# because we doesn't want to have too many updates
```
bmuenzenmeyer, line 2:
```suggestion
# because we don't want to have too many updates
```

bmuenzenmeyer (approved):

PR #191, opened 2025-02-17
feat: create linter system

### LINKED ISSUE #56: Introduce a Linter / Validation System
This is another feature we want to tackle and have within the API docs tooling. The idea is similar to ESlint: We have a system that reports on irregularities when running the API docs tooling against files.

Assume you have a missing version on one of the YAML metadata (i.e.: `version: REPLACEME`), or some of the YAML is missing (like each module may have an "introduced_in" on the first node/section), or you have some invalid type within the API docs (ie: `{SomethingThatDoesNotExist}` or a typo `{Obrect}` instead of `{Object}`

- The system would print on the console with the error level (info, warn, error) and tell in which line and position it would be such an issue on the source Markdown file.
- The system could be a singleton that collects all issues during runtime and then displays them at the end of the run.
- The system could be disabled/ignored with a CLI flag `--skip-validation`;
The system must know the almost exact position of something, possibly within the AST tree, which gives you a position object for each item so you can narrow it to a close position.
- Some methods might need to be updated (if they don't use a node (ApiMetadataEntry or Parent/Node) to pass such information or the position object; Or maybe some abstraction so that the `queries.mjs`, `generators.mjs` and all utils know what is the current context being handled. (Note that since parsing and processing is async, we cannot set a static context within a Singleton)
The system should not use try/catch or exceptions since the code should be evergreen and failproof to actual errors within the markdown.
- All the message templates should be defined within the `constants.mjs`
- When there's at least one Error level, the process will exit with a non-success status code.

### Possible items for the system:

- Warning: missing version on a YAML metadata entry (changes or updates properties)
- Warning: invalid version on a YAML metadata entry (changes or updates properties) (ie: `REPLACEME`)
- Warning: invalid type reference {type}
- Info: missing "introduced_in" on top of the document
- More?

araujogui:
## Description

Create linter system

## Validation

Run CLI 

## Related Issues

Fixes #56

### Check List


bin/cli.mjs
```diff
@@ -66,7 +74,17 @@ program
...
  * @description The return type for values sent to the program from the CLI.
  */
-const { input, output, target = [], version, changelog } = program.opts();
+const {
+  input,
+  output,
+  target = [],
+  version,
+  changelog,
+  skipValidation,
```
flakey5, line 83:
```suggestion
  skipLinting,
```
same for below uses

src/linter/index.mjs
```diff
@@ -0,0 +1,66 @@
+// @ts-check
```
flakey5, line 1:
```suggestion
```

src/linter/reporters/console.mjs
```diff
@@ -0,0 +1,31 @@
+// @ts-check
```
flakey5, line 1:
```suggestion
```

src/linter/reporters/github.mjs
```diff
@@ -0,0 +1,24 @@
+// @ts-check
```
flakey5, line 1:
```suggestion
```

src/linter/reporters/github.mjs
```diff
@@ -0,0 +1,24 @@
...
+ * @type {import('../types.d.ts').Reporter}
+ */
+export default issue => {
+  const actions = {
+    warn: core.warning,
+    error: core.error,
+    info: core.notice,
+  };
+
+  (actions[issue.level] || core.notice)(issue.message, {
```
flakey5, line 19:
```suggestion
  const logFn = actions[issue.level] || core.notice

  logFn(issue.message, {
```

nit

flakey5 (commented):
first pass will finish in a bit

src/constants.mjs
```diff
@@ -125,17 +125,45 @@ export const DOC_SLUG_ENVIRONMENT = 'environment-variables-1';
...
-    'EvalError', 'Function', 'Map', 'NaN', 'Object', 'Promise', 'Proxy', 'RangeError',
-    'ReferenceError', 'RegExp', 'Set', 'SharedArrayBuffer', 'SyntaxError', 'Symbol',
-    'TypeError', 'URIError', 'WeakMap', 'WeakSet',
-
-    'TypedArray',
-    'Float32Array', 'Float64Array',
-    'Int8Array', 'Int16Array', 'Int32Array',
-    'Uint8Array', 'Uint8ClampedArray', 'Uint16Array', 'Uint32Array',
-  ].map(e => [e, e])),
+  ...Object.fromEntries(
```
AugustinMauroy, line 128: why theses changes ?

AugustinMauroy:
do we have ability to warn if it's an unknown stability level ?

src/constants.mjs
araujogui: Just prettier's auto formatting changes

araujogui:
> do we have ability to warn if it's an unknown stability level ?

I didn't create this rule but it's a good idea, I can develop this.

src/constants.mjs
AugustinMauroy: oh sad

flakey5 (approved):

flakey5:
Wdyt of adding a capability here to ignore only specific linting errors? I.e. being able to ignore an invalid `introduced_in` but still getting errors on everything else

AugustinMauroy:
+1 for ingnoring rules

bin/cli.mjs
```diff
@@ -12,6 +12,8 @@ import generators from '../src/generators/index.mjs';
 import createMarkdownLoader from '../src/loaders/markdown.mjs';
 import createMarkdownParser from '../src/parsers/markdown.mjs';
 import createNodeReleases from '../src/releases.mjs';
+import { Linter } from '../src/linter/index.mjs';
```
ovflowd, line 15: I'd recommend following the same approach of "createLinter"

bin/cli.mjs
```diff
@@ -61,12 +69,24 @@ program
...
+  input,
+  output,
+  target = [],
+  version,
+  changelog,
+  skipLinting,
+  reporter,
+} = program.opts();
+
+const linter = skipLinting ? undefined : new Linter();
```
ovflowd, line 89: Maybe always create the Linter instance but allow it to run on a dry-run mode?

bin/cli.mjs
```diff
@@ -92,3 +114,11 @@ await runGenerators({
   // A list of all Node.js major versions with LTS status
   releases: await getAllMajors(),
 });
+
+if (linter) {
```
ovflowd, line 118:
```ts

linter.report() -- if dry run does nothing maybe

exit(Boolean(linter.hasError))
```

package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "acorn": "^8.14.0",
+    "@actions/core": "^1.11.1",
```
ovflowd, line 33: Why we need actions as a dep?

src/constants.mjs
ovflowd: Is this the project's prettier settings or yours?

package.json
MattIPv4: For src/linter/reporters/github.mjs I would guess?

src/constants.mjs
araujogui: Project
image

package.json
araujogui: Yep

src/generators/types.d.ts
```diff
@@ -30,6 +31,8 @@ declare global {
 
     // A list of all Node.js major versions and their respective release information
     releases: Array<ApiDocReleaseEntry>;
+
+    linter: Linter | undefined;
```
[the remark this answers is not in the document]
araujogui: Mb

src/linter/utils/semver.mjs
```diff
@@ -0,0 +1,13 @@
+/**
+ * Validates a semver version string
+ *
+ * @param {string} version
+ * @returns {boolean}
+ */
+export const validateVersion = version => {
+  // https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+  const regex =
```
ovflowd, line 9: We could use `semver` package, instead of maintaining our own regex. I do recall we already use semver package here? And it has a method to validate if a string is a semver.

src/types.d.ts
```diff
@@ -99,6 +99,7 @@ declare global {
     // Extra YAML section entries that are stringd and serve
     // to provide additional metadata about the API doc entry
     tags: Array<string>;
+    yaml_position: Position;
```
ovflowd, line 102: Can you add a documentation (inline) of what this exactly is and for what it serves?

ovflowd (approved):
Such a clean code! LGTM! ✨

AugustinMauroy (approved):
If it's not too complicated have the rules deactivated if not LGTM

bjohansebas (approved):
LGTM, is there any way to test this?

AugustinMauroy:
> LGTM, is there any way to test this?

clone this repo `npx api-docs-tooling -i "/path/to/node/doc/api/*.md"`

src/linter/reporters/console.mjs
```diff
@@ -0,0 +1,31 @@
...
+ * @type {import('../types.d.ts').Reporter}
+ */
+export default issue => {
+  const position = issue.location.position
+    ? ` (${issue.location.position.start.line}:${issue.location.position.end.line})`
+    : '';
+
+  console.log(
+    styleText(
+      // @ts-expect-error ForegroundColors is not exported
```
canerakdas, line 26:
```suggestion
```

bjohansebas:
> > LGTM, is there any way to test this?
>
> clone this repo npx api-docs-tooling -i "/path/to/node/doc/api/*.md"

I was referring to more automated tests without having to take that step you mentioned.

ovflowd:
Agreed some unit tests would be cool

araujogui:
Can I merge this PR and create a follow-up pr with unit tests, a rule deactivation?

ovflowd (approved):
Can we add some unit tests, @araujogui?

bjohansebas (approved):
LGTM!

araujogui:
GitHub seems to be automatically stripping ANSI escape sequences, so tests pass locally and fail in CI

src/linter/tests/fixtures/entries.mjs
```diff
@@ -0,0 +1,84 @@
...
+ * Noop function.
+ *
+ * @returns {any}
+ */
+const noop = () => {};
+
+/**
+ * @type {ApiDocMetadataEntry}
+ */
+export const assertEntry = {
```
ovflowd, line 11: Can't this be a json file?

ovflowd (approved):
LGTM

src/linter/tests/fixtures/entries.mjs
araujogui: It has two mocked functions.

araujogui:
> > GitHub seems to be automatically stripping ANSI escape sequences, so tests pass locally and fail in CI
> 
> Any workaround?

Yep, FORCE_COLOR=1 fixed the issue.

PR #209, opened 2025-03-14
Remove assumption regarding local git clone

aduh95:
IMO it's bad practice to assume the doc will be in a git directory, and to assume it would have an `origin` remote pointing to HTTPS or SSH GitHub repo. It would be much better to let the user opt-in into that behavior, and default to an hard-coded value.

### Check List

### DIFF
--- a/src/generators/api-links/index.mjs
+++ b/src/generators/api-links/index.mjs
@@ -54,7 +54,7 @@ export default {
     if (input.length > 0) {
       const repositoryDirectory = dirname(input[0].path);
 
-      const repository = getBaseGitHubUrl(repositoryDirectory);
+      const repository = 'https://github.com/nodejs/node';
 
       const tag = getCurrentGitHash(repositoryDirectory);
 
--- a/src/generators/api-links/index.mjs
+++ b/src/generators/api-links/index.mjs
@@ -2,10 +2,6 @@
 
 import { basename, dirname, join } from 'node:path';
 import { writeFile } from 'node:fs/promises';
-import {
-  getBaseGitHubUrl,
-  getCurrentGitHash,
-} from './utils/getBaseGitHubUrl.mjs';
 import { extractExports } from './utils/extractExports.mjs';
 import { findDefinitions } from './utils/findDefinitions.mjs';
 import { checkIndirectReferences } from './utils/checkIndirectReferences.mjs';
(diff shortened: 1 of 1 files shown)


src/generators/api-links/index.mjs
```diff
@@ -2,10 +2,6 @@
 
 import { basename, dirname, join } from 'node:path';
 import { writeFile } from 'node:fs/promises';
-import {
```
ovflowd, line 5: Can you delete these utils if they're not used anywhere else?

src/generators/api-links/index.mjs
```diff
@@ -54,9 +50,9 @@ export default {
     if (input.length > 0) {
       const repositoryDirectory = dirname(input[0].path);
 
-      const repository = getBaseGitHubUrl(repositoryDirectory);
+      const repository = 'https://github.com/nodejs/node';
 
-      const tag = getCurrentGitHash(repositoryDirectory);
+      const tag = 'HEAD';
```
ovflowd, line 55: Is it fine for us to assume HEAD?

AugustinMauroy (approved):

src/generators/api-links/index.mjs
flakey5: In dev it shouldn't really matter, but for releases it needs to be a hash since the line numbers referenced are going to change eventually.

I.e. for the latest v20 release, https://github.com/nodejs/node/blob/bb1a61d8737feff534bb85368dab3b7c554c863d/lib/_http_client.js#L869 is correct vs https://github.com/nodejs/node/blob/HEAD/lib/_http_client.js#L869 points to something else.

flakey5:
Fwiw the current behavior is almost directly taken from the old doc tooling https://github.com/nodejs/node/blob/63e3cc7e82a645b0763174f306bfe589164ce7d6/tools/doc/apilinks.mjs#L38-L45

flakey5 (changes requested):
> It would be much better to let the user opt-in into that behavior

How will a user be able to opt-in?

aduh95:
> > It would be much better to let the user opt-in into that behavior
> 
> How will a user be able to opt-in?

There are numerous possibilities (env variables, CLI flag, some config file). Anyway, my point is that it’s better IMO to

1. Have it opt-in, rather than
3. not have it at all, rather than
4. Have is opt-out, rather than
5. Have it without any way for disabling it

Currently, this PR is not implementing the best solution, but I still think it’s an improvement

ovflowd:
I think the PR as it stands would break current behaviour. Let's not do that. I agree with opt-in, but then this PR needs to be updated to be able to do that. The changes should be _relatively_ simple IMO.

PR #212, opened 2025-03-17
linter: add deprecation code check, linter declarations

flakey5:
Adds a check to the linter that ensures all of the deprecation codes in `doc/api/deprecations.md` are in numerical order.

This also required adding linting declarations to the linter (i.e. `<!-- md-lint ... -->`), since some of the missing deprecations are ignored for whatever reason.

TODO:
- [ ] write tests for the check & declarations
- [ ] deprecation type check
- [ ] cleanup

### DIFF
--- a/src/linter/engine.mjs
+++ b/src/linter/engine.mjs
@@ -10,13 +10,15 @@ const createLinterEngine = rules => {
    * Validates a ApiDocMetadataEntry entry against all defined rules
    *
    * @param {ApiDocMetadataEntry} entry
+   * @param {import('./types').LintDeclarations}
+   * @param declarations
    * @returns {import('./types').LintIssue[]}
    */
-  const lint = entry => {
+  const lint = (entry, declarations) => {
     const issues = [];
 
     for (const rule of rules) {
-      const ruleIssues = rule(entry);
+      const ruleIssues = rule([entry], declarations);
 
       if (ruleIssues.length > 0) {
         issues.push(...ruleIssues);
@@ -30,13 +32,19 @@ const createLinterEngine = rules => {
    * Validates an array of ApiDocMetadataEntry entries against all defined rules
    *
    * @param {ApiDocMetadataEntry[]} entries
+   * @param {import('./types').LintDeclarations}
+   * @param declarations
    * @returns {import('./types').LintIssue[]}
    */
-  const lintAll = entries => {
+  const lintAll = (entries, declarations) => {
     const issues = [];
 
-    for (const entry of entries) {
-      issues.push(...lint(entry));
+    for (const rule of rules) {
+      const ruleIssues = rule(entries, declarations);
+
+      if (ruleIssues.length > 0) {
+        issues.push(...ruleIssues);
+      }
     }
 
     return issues;
--- a/src/linter/index.mjs
+++ b/src/linter/index.mjs
@@ -1,5 +1,6 @@
 'use strict';
 
+import { LINT_MESSAGES } from '../constants.mjs';
 import createLinterEngine from './engine.mjs';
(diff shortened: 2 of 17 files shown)


flakey5:
cc @ovflowd @araujogui

src/linter/engine.mjs
```diff
@@ -10,13 +10,15 @@ const createLinterEngine = rules => {
    * Validates a ApiDocMetadataEntry entry against all defined rules
    *
    * @param {ApiDocMetadataEntry} entry
+   * @param {import('./types').LintDeclarations}
+   * @param declarations
    * @returns {import('./types').LintIssue[]}
    */
-  const lint = entry => {
+  const lint = (entry, declarations) => {
```
flakey5, line 17: The declarations kinda feel retrofitted in but not sure if there's a better way to pass them

src/linter/index.mjs
```diff
@@ -58,6 +66,87 @@ const createLinter = (dryRun, disabledRules) => {
...
+   */
+  const parseLinterDeclaration = declaration => {
+    // Trim off any excess spaces from the beginning & end
+    declaration = declaration.trim();
+
+    // Extract the name for the declaration
+    const [name, ...value] = declaration.split(' ');
+
+    switch (name) {
+      case 'skip-deprecation': {
```
flakey5, line 82: Wdyt of defining these definitions similarly to the rules & reporters? (i.e. have them defined in a `declarations/skip-deprecation.mjs`, then here we look them up & call them if they exist)

src/linter/rules/deprecation-code-order.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+  const hierarchy = buildHierarchy(entries);
+
+  hierarchy.forEach(root => {
+    const deprecations = getDeprecationEntries(root.hierarchyChildren);
+
+    let expectedCode = 1;
+
+    for (const deprecation of deprecations || []) {
+      while (declarations.skipDeprecation.includes(expectedCode)) {
```
flakey5, line 71: probably a nicer way to do this loop

AugustinMauroy:
This [test](https://github.com/nodejs/api-docs-tooling/blob/222d8b2bcde5b51b11f5facb3e93798de3284e72/src/linter/tests/engine.test.mjs) fail because it's use [this fixture](
 https://github.com/nodejs/api-docs-tooling/blob/222d8b2bcde5b51b11f5facb3e93798de3284e72/src/linter/tests/fixtures/entries.mjs)
 
 Maybe use snapshot instead of fixture 

if you use snapshot you ill need:
```json
    "test": "node --test --experimental-test-snapshots",
    "test:coverage": "node --experimental-test-coverage --test",
    "test:update-snapshots": "node --test --experimental-test-snapshots --test-update-snapshots",
    "test:watch": "node --test --experimental-test-snapshots --watch",
```

and test will look like:

```js
it('should call each rule with the provided entry', (t) => {
    const rule1 = mock.fn(() => []);
    const rule2 = mock.fn(() => []);

    const engine = createLinterEngine([rule1, rule2]);

    engine.lint(assertEntry);

    assert.strictEqual(rule1.mock.callCount(), 1);
    assert.strictEqual(rule2.mock.callCount(), 1);

    t.assert.snapshot(rule1.mock.calls[0].arguments);
    t.assert.snapshot(rule2.mock.calls[0].arguments);
  });
  ```

src/linter/engine.mjs
araujogui: I think declarations could be part of metadata entries

avivkeller:
@flakey5 has this stale-ed?

PR #215, opened 2025-03-21
feat: update stability css

avivkeller:
This adds the styles that @aduh95 is adding in https://github.com/nodejs/node/pull/57516.

As a TODO: A validator for duplicate stability nodes should be added to the Linter / Validation system.

(Note: I've added the "Blocked" label, as the original PR is yet to land)

### DIFF
--- a/src/generators/legacy-html/assets/style.css
+++ b/src/generators/legacy-html/assets/style.css
@@ -318,6 +318,10 @@ hr.line {
   margin: 0 0 1rem;
   padding: 1rem;
   line-height: 1.5;
+
+  position: sticky;
+  top: 3rem;
+  z-index: 1;
 }
 
 .api_stability p:first-of-type {
@@ -867,6 +871,10 @@ kbd {
 @media not screen, (width <= 600px) {
   .header {
     position: relative;
+    top: 0;
+  }
+
+  .api_stability {
     top: 0;
   }
 }


src/linter/engine.mjs
```diff
@@ -3,9 +3,12 @@
 /**
  * Creates a linter engine instance to validate ApiDocMetadataEntry entries
  *
- * @param {import('./types').LintRule} rules Lint rules to validate the entries against
+ * @param {{
+ * multiEntryRules: import('./types').MultipleEntriesLintRules[]
```
araujogui, line 7: I think all rules could be a multi-entry rule, the singleEntryRules can just loop all entries.

avivkeller: That makes a lot more sense than what I had setup, I'll adjust it to that, thanks!

avivkeller:
@nodejs/web-infra @araujogui please (re-)review the lint changes.

Thank you!

src/linter/rules/duplicate-stability-nodes.mjs
```diff
@@ -0,0 +1,38 @@
...
+      depth > currentDepth &&
+      entryStability >= 0 &&
+      entryStability === currentStability
+    ) {
+      issues.push({
+        level: 'warn',
+        message: LINT_MESSAGES.duplicateStabilityNode,
+        location: {
+          path: entry.api_doc_source,
+          position: entry.yaml_position,
```
avivkeller, line 28: This isn't the _exact_ position of the stability node, but it's right around where the YAML is.

Is there a better position to use?

AugustinMauroy (approved):

flakey5 (approved):

avivkeller:
I changed the logic a little bit, please re-review

src/linter/tests/rules/invalid-change-version.test.mjs
```diff
@@ -5,16 +5,21 @@ import { assertEntry } from '../fixtures/entries.mjs';
...
-    const issues = invalidChangeVersion({
-      ...assertEntry,
-      changes: [...assertEntry.changes, { version: ['v13.9.0', 'REPLACEME'] }],
-    });
+    const issues = invalidChangeVersion([
+      {
+        ...assertEntry,
+        changes: [
+          ...assertEntry.changes,
+          { version: ['v13.9.0', 'REPLACEME'] },
```
aduh95, line 19: Not related to this PR, but FWIW it's unrealistic to have a specific version AND `REPLACEME`, either a feature has been released (and `REPLACEME` has been replaced), either it has not been released yet

avivkeller: That's for testing reasons, and even so, it's being changed in #238

PR #230, opened 2025-03-23
fix: followup globSync fix

avivkeller:
# 😅 

I missed the `javascript.mjs` file. (And while I'm following up, I should've used the `node:` prefix)


ovflowd (approved):
👀 

#shame xD

PR #233, opened 2025-03-24
feat: display og images in social media

bjohansebas:
With this, the images will be displayed on social media. It's not very important, but it's a small detail

https://nodejs.org/en/next-data/og/announcement/All%20%7C%20Node.js%20v22.14.0%20Documentation

output: 
image

### DIFF
--- a/src/generators/legacy-html-all/index.mjs
+++ b/src/generators/legacy-html-all/index.mjs
@@ -85,7 +85,7 @@ export default {
     const generatedAllTemplate = apiTemplate
       .replace('__ID__', 'all')
       .replace(/__FILENAME__/g, 'all')
-      .replace('__SECTION__', 'All')
+      .replace(/__SECTION__/g, 'All')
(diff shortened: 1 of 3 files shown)


flakey5 (approved):

AugustinMauroy (changes requested):
LGMT ! But we should have ability to op-out OG image. Because `nodejs/nodejs.org` can be output as static so OG will be 404

avivkeller (approved):
LGTM

bjohansebas:
> But we should have ability to op-out OG image. Because nodejs/nodejs.org can be output as static so OG will be 404

Any idea how to solve that? Also, if that starts returning 404, then the OG image wouldn't work on **nodejs/nodejs.org** either.

ovflowd (changes requested):
I don't think we should make any changes to the legacy generators.

avivkeller:
Given the comment above, I've closed this. However, you can always re open to continue discussion

PR #237, opened 2025-03-24
revert: undo globSync from fs

avivkeller:
CC @ovflowd

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "dedent": "^1.5.3",
         "estree-util-visit": "^2.0.0",
         "github-slugger": "^2.0.0",
+        "glob": "^11.0.1",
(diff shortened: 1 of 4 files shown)


bjohansebas (approved):

avivkeller:
I think you are thinking of #235, which was a different PR

ovflowd:
What I recall is that the original PR has something “windows” on the PR title but I might be mistaken

ovflowd (approved):

avivkeller:
FWIW #228 (The original PR) stated "On Windows, using a backslash in the path (e.g., -i .\docs\buffer.md rather than -i ./docs/buffer.md) prevents the glob from matching correctly. This issue is resolved by using the fs version of globSync." in the description

PR #238, opened 2025-03-24
chore(lint): report error on invalid change versions, except REPLACEME

avivkeller:
Per https://github.com/nodejs/node/pull/57343#pullrequestreview-2711832418
> * We should not report REPLACEME as an invalid version.
> * We should report invalid versions as a hard failure.

### DIFF
--- a/src/linter/rules/invalid-change-version.mjs
+++ b/src/linter/rules/invalid-change-version.mjs
@@ -1,42 +1,38 @@
 import { LINT_MESSAGES } from '../constants.mjs';
 import { valid } from 'semver';
+import { env } from 'node:process';
+
+const NODE_RELEASED_VERSIONS = env.NODE_RELEASED_VERSIONS?.split(',');
 
 /**
- * Checks if any change version is invalid
+ * Determines if a given version is invalid.
  *
- * @param {ApiDocMetadataEntry[]} entries
- * @returns {Array<import('../types').LintIssue>}
+ * @param {string} version - The version to check.
+ * @returns {boolean} True if the version is invalid, false otherwise.
  */
-export const invalidChangeVersion = entries => {
-  const issues = [];
+const isInvalid = NODE_RELEASED_VERSIONS
+  ? version =>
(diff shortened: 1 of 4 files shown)


AugustinMauroy (approved):

src/linter/tests/rules/invalid-change-version.test.mjs
```diff
@@ -13,12 +13,15 @@ describe('invalidChangeVersion', () => {
   it('should return an issue if a change version is invalid', () => {
     const issues = invalidChangeVersion({
       ...assertEntry,
-      changes: [...assertEntry.changes, { version: ['v13.9.0', 'REPLACEME'] }],
+      changes: [
+        ...assertEntry.changes,
+        { version: ['v13.9.0', 'INVALID_VERSION'] },
```
aduh95, line 18: I would keep the test case to ensure we report as an error if `REPLACEME` is not the only version
```suggestion
        { version: ['v13.9.0', 'REPLACEME'] },
```

src/linter/rules/invalid-change-version.mjs
```diff
@@ -19,11 +19,11 @@ export const invalidChangeVersion = entry => {
     );
 
   const invalidVersions = allVersions.filter(
-    version => valid(version) === null
+    version => version !== 'REPLACEME' && valid(version) === null
```
aduh95, line 22: I would filter out `REPLACEME` before reaching here, and we need to support passing `NODE_RELEASED_VERSIONS` via the env
```suggestion
    process.env.NODE_RELEASED_VERSIONS ?
      version => process.env.NODE_RELEASED_VERSIONS.includes(version) :
      version => !valid(version),
```

avivkeller:
I've updated this with support for `NODE_RELEASED_VERSIONS`.

src/linter/tests/rules/invalid-change-version.test.mjs
avivkeller: Got it! Thanks for the clarification

aduh95 (commented):
LGTM

ovflowd (approved):
LGTM

PR #239, opened 2025-03-28
feat(linter): introduce raw ast rules

araujogui:
### Check List

### DIFF
--- a/src/linter/engine.mjs
+++ b/src/linter/engine.mjs
@@ -10,14 +10,14 @@ const createLinterEngine = rules => {
    * Validates an array of mdast trees against all defined rules
    *
    * @param {import('vfile').VFile} file
-   * @param {import('mdast').Root[]} ast
+   * @param {import('mdast').Root[]} tree
    * @returns {import('./types').LintIssue[]}
    */
-  const lint = (file, ast) => {
+  const lint = (file, tree) => {
     const issues = [];
 
     for (const rule of rules) {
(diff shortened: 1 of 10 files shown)


src/linter/engine.mjs
```diff
@@ -1,29 +1,29 @@
...
+   * @param {import('mdast').Root[]} ast
    * @returns {import('./types').LintIssue[]}
    */
-  const lintAll = entries => {
+  const lint = ast => {
     const issues = [];
 
     for (const rule of rules) {
-      issues.push(...rule(entries));
+      issues.push(...rule(ast));
```
avivkeller, line 19: WDYT about there being be a way for the linter to accept raw OR ast, depending on what it needs. For some scenarios, getting the AST may be preferable

araujogui: I don't have a strong opinion about this, I can make a simple test so we can decide what is the best option.

araujogui:
I refactored the `invalid-change-version` to an initial version that uses mdast and yaml ast to correctly report line numbers.   Feedback is welcome @nodejs/web-infra

AugustinMauroy (approved):

avivkeller (approved):

src/linter/engine.mjs
avivkeller: Giving another look, and I think it's fine like this. Looks really good :-)

araujogui:
I'll open another PR introducing this.

PR #246, opened 2025-04-06
fix(lint): hardcode ignored invalid change versions

avivkeller:
Ref: https://github.com/nodejs/node/pull/57343#issuecomment-2781707538

### DIFF
--- a/src/linter/rules/invalid-change-version.mjs
+++ b/src/linter/rules/invalid-change-version.mjs
@@ -18,6 +18,9 @@ const isValidReplaceMe = (version, length) =>
  * Checks if a given semantic version should be ignored.
  * A version is considered ignored if its major version is 0 and minor version is less than 2.
  *
+ * These versions are extremely old, and are not shown in the changelog used to generate
+ * `NODE_RELEASED_VERSIONS`, so they must be hardcoded.
+ *
  * @param {string} version - The version to check.
  * @returns {boolean} Returns true if the version is ignored, false otherwise.
  */
--- a/src/linter/rules/invalid-change-version.mjs
+++ b/src/linter/rules/invalid-change-version.mjs
@@ -22,7 +22,6 @@ const isValidReplaceMe = (version, length) =>
  * @returns {boolean} Returns true if the version is ignored, false otherwise.
  */
 const isIgnoredVersion = version => {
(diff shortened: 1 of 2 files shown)


src/linter/rules/invalid-change-version.mjs
```diff
@@ -14,6 +14,19 @@ const NODE_RELEASED_VERSIONS = env.NODE_RELEASED_VERSIONS?.split(',');
...
+/**
+ * Checks if a given semantic version should be ignored.
+ * A version is considered ignored if its major version is 0 and minor version is less than 2.
+ *
+ * @param {string} version - The version to check.
+ * @returns {boolean|undefined} Returns true if the version is ignored, false otherwise.
+ */
+const isIgnoredVersion = version => {
+  if (version === 'v0.11.15') return true;
+  const { major, minor } = parse(version) || {};
```
avivkeller, line 26: Note: We are intentionally not throwing on an invalid semver, as `NODE_RELEASED_VERSIONS` has the final say on what's a valid version (although, a non-semver valid version would never happen)

avivkeller:
Also see https://github.com/nodejs/node/pull/57779

bjohansebas (approved):

src/linter/tests/fixtures/invalidChangeVersion-environment.mjs
```diff
@@ -7,7 +7,7 @@ const issues = invalidChangeVersion([
     ...assertEntry,
     changes: [
       ...assertEntry.changes,
-      { version: ['SOME_OTHER_RELEASED_VERSION'] },
+      { version: ['SOME_OTHER_RELEASED_VERSION', 'v0.1.2'] },
```
AugustinMauroy, line 10: maybe add comment to explain why this version is hardcoded

araujogui (approved):

src/linter/tests/fixtures/invalidChangeVersion-environment.mjs
avivkeller: @aduh95 I'm assuming the reason is they are not included in the changelog, just confirming with you?

AugustinMauroy (approved):

ovflowd (approved):

PR #251, opened 2025-04-16
fix(changelog): allow for non-https protocols

avivkeller:
## Description

We should only get the file from the file system when we're certain the URL is referencing a file. This prevents misinterpreting other protocols—like `http` instead of `https`—as file paths.

## Validation

See description

### Check List


flakey5 (approved):

AugustinMauroy (approved):

ovflowd (approved):
LGTM

bjohansebas (approved):

PR #253, opened 2025-04-20
feat(generator): add `node.config.json`

### LINKED ISSUE #216: JSON Config Schema Generator

Re nodejs/node#57016 nodejs/node#57171

There is a new generator for `node.config.json` schema. The doc tooling just calls an internal function and writes what it returns to a file. The internal function (`generateJsonSchema`) could be moved into a new generator here.

However, it's worth noting that that function calls a native function (`getEnvOptionsInputType`) that does need to stay in Node core.

avivkeller:
## Description

This generates the https://nodejs.org/node-config-schema.json file (https://raw.githubusercontent.com/nodejs/node/main/doc/node-config-schema.json).

## Validation

Other than this schema including `experimental-quic` (which the Node core one does not), it should be a 1:1 match with the generator in Node core (This is likely due to the machine generating the node core schema not having quic).

[Example Output](https://aviv.sh/hostables/nodejs/1745190326-node-config-schema.json)

## Related Issues

Fixes #216

### Check List

### DIFF
--- a/src/generators/node-config-schema/constants.mjs
+++ b/src/generators/node-config-schema/constants.mjs
@@ -3,9 +3,8 @@ export const ERRORS = {
   missingCCandHFiles:
     'Both node_options.cc and node_options.h must be provided.',
   headerTypeNotFound:
-    'Header type for "{{headerKey}}" not found in the header file.',
-  missingTypeDefinition:
-    'No type definition found for header type "{{headerType}}" in TYPE_DEFINITION_MAP.',
+    'A type for "{{headerKey}}" not found in the header file.',
+  missingTypeDefinition: 'No type schema found for "{{type}}".',
 };
 
 // Regex pattern to match calls to the AddOption function.
@@ -14,38 +13,3 @@ export const ADD_OPTION_REGEX =
 
 // Regex pattern to match header keys in the Options class.
 export const OPTION_HEADER_KEY_REGEX = /Options::(\w+)/;
-
-// Basic JSON schema for node.config.json
-export const BASIC_SCHEMA = {
-  $schema: 'https://json-schema.org/draft/2020-12/schema',
-  additionalProperties: false,
-  properties: {
-    $schema: {
-      type: 'string',
-    },
-    nodeOptions: {
-      additionalProperties: false,
-      properties: {},
-      type: 'object',
-    },
-  },
-  type: 'object',
-};
-
-// Schema Definition Map for Data Types
-export const TYPE_DEFINITION_MAP = {
-  'std::vector<std::string>': {
-    oneOf: [
-      { type: 'string' }, // Single string case
-      {
-        items: { type: 'string', minItems: 1 }, // Array of strings, ensuring at least one item
-        type: 'array',
-      },
-    ],
-  },
-  uint64_t: { type: 'number' }, // 64-bit unsigned integer maps to a number
-  int64_t: { type: 'number' }, // 64-bit signed integer maps to a number
-  HostPort: { type: 'number' }, // HostPort is a number, like 4000
-  'std::string': { type: 'string' }, // String type
-  bool: { type: 'boolean' }, // Boolean type
-};
--- a/src/generators/node-config-schema/constants.mjs
+++ b/src/generators/node-config-schema/constants.mjs
@@ -1,0 +1,51 @@
+// Error Messages
+export const ERRORS = {
+  missingCCandHFiles:
+    'Both node_options.cc and node_options.h must be provided.',
+  headerTypeNotFound:
+    'Header type for "{{headerKey}}" not found in the header file.',
+  missingTypeDefinition:
+    'No type definition found for header type "{{headerType}}" in TYPE_DEFINITION_MAP.',
+};
+
+// Regex pattern to match calls to the AddOption function.
+export const ADD_OPTION_REGEX =
+  /AddOption[\s\n\r]*\([\s\n\r]*"([^"]+)"(.*?)\);/gs;
+
+// Regex pattern to match header keys in the Options class.
+export const OPTION_HEADER_KEY_REGEX = /Options::(\w+)/;
+
+// Basic JSON schema for node.config.json
+export const BASIC_SCHEMA = {
+  $schema: 'https://json-schema.org/draft/2020-12/schema',
+  additionalProperties: false,
+  properties: {
+    $schema: {
+      type: 'string',
+    },
+    nodeOptions: {
+      additionalProperties: false,
+      properties: {},
+      type: 'object',
+    },
+  },
+  type: 'object',
+};
+
+// Schema Definition Map for Data Types
+export const TYPE_DEFINITION_MAP = {
+  'std::vector<std::string>': {
+    oneOf: [
+      { type: 'string' }, // Single string case
+      {
+        items: { type: 'string', minItems: 1 }, // Array of strings, ensuring at least one item
+        type: 'array',
+      },
+    ],
+  },
+  uint64_t: { type: 'number' }, // 64-bit unsigned integer maps to a number
+  int64_t: { type: 'number' }, // 64-bit signed integer maps to a number
+  HostPort: { type: 'number' }, // HostPort is a number, like 4000
+  'std::string': { type: 'string' }, // String type
+  bool: { type: 'boolean' }, // Boolean type
+};
--- a/src/generators/node-config-schema/index.mjs
+++ b/src/generators/node-config-schema/index.mjs
@@ -1,11 +1,13 @@
 import { readFile, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+
 import {
-  ERRORS,
   ADD_OPTION_REGEX,
+  ERRORS,
   OPTION_HEADER_KEY_REGEX,
 } from './constants.mjs';
-import { join } from 'node:path';
 import schema from './schema.json' with { type: 'json' };
+import { formatErrorMessage, getTypeSchema } from './utilities.mjs';
 
 /**
  * This generator generates the `node.config.json` schema.
@@ -55,6 +57,7 @@ export default {
       }
 
       const headerKey = config.match(OPTION_HEADER_KEY_REGEX)?.[1];
+
       // If there's no header key, it's either a V8 option or a no-op
       if (!headerKey) {
         continue;
@@ -91,47 +94,3 @@ export default {
     return schema;
   },
 };
-
-/**
- * Helper function to replace placeholders in error messages with dynamic values.
- * @param {string} message - The error message with placeholders.
- * @param {Object} params - The values to replace the placeholders.
- * @returns {string} - The formatted error message.
- */
-function formatErrorMessage(message, params) {
-  return message.replace(/{{(\w+)}}/g, (_, key) => params[key] || `{{${key}}}`);
-}
-
-/**
- * Returns the JSON Schema definition for a given C++ type.
- *
- * @param {string} type - The type to get the schema for.
- * @returns {object} JSON Schema definition for the given type.
- */
-function getTypeSchema(type) {
-  switch (type) {
-    case 'std::vector<std::string>':
-      return {
-        oneOf: [
-          { type: 'string' },
-          {
-            type: 'array',
-            items: { type: 'string' },
-            minItems: 1,
-          },
-        ],
-      };
-    case 'uint64_t':
-    case 'int64_t':
-    case 'HostPort':
-      return { type: 'number' };
-    case 'std::string':
-      return { type: 'string' };
-    case 'bool':
-      return { type: 'boolean' };
-    default:
-      throw new Error(
-        formatErrorMessage(ERRORS.missingTypeDefinition, { type })
-      );
-  }
-}
--- a/src/generators/node-config-schema/index.mjs
+++ b/src/generators/node-config-schema/index.mjs
@@ -5,7 +5,7 @@ import {
   OPTION_HEADER_KEY_REGEX,
 } from './constants.mjs';
 import { join } from 'node:path';
-import schema from './schema.json' with { type: 'json ' };
+import schema from './schema.json' with { type: 'json' };
 
 /**
  * This generator generates the `node.config.json` schema.
--- a/src/generators/node-config-schema/index.mjs
+++ b/src/generators/node-config-schema/index.mjs
@@ -5,6 +5,7 @@ import {
   OPTION_HEADER_KEY_REGEX,
 } from './constants.mjs';
 import { join } from 'node:path';
+import schema from './schema.json' with { type: 'json ' };
 
 /**
  * This generator generates the `node.config.json` schema.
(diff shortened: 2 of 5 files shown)


avivkeller:
ack! This contains a commit from #252

src/generators/node-config-schema/constants.mjs
```diff
@@ -0,0 +1,51 @@
...
+
+// Regex pattern to match calls to the AddOption function.
+export const ADD_OPTION_REGEX =
+  /AddOption[\s\n\r]*\([\s\n\r]*"([^"]+)"(.*?)\);/gs;
+
+// Regex pattern to match header keys in the Options class.
+export const OPTION_HEADER_KEY_REGEX = /Options::(\w+)/;
+
+// Basic JSON schema for node.config.json
+export const BASIC_SCHEMA = {
```
AugustinMauroy, line 19: Is this constant should be external ? I mean is it can be pass as arg.

Because now if there change on core we need to update this repo. And ideally we want to not touch to this repo

ovflowd, line 19: This imo should be a JSON file.

avivkeller: It's extremely rare that these types change. The last time a new type was introduced was 7 years ago (https://github.com/nodejs/node/commit/29a71bae40ffa0bbc8ba6b2bdf051a09987da7f7).

If and when a new type is added, it's just a few lines of code to update this repository.

flakey5: I think the json schema should be put in this repository since there isn't really anything necessitating it to be in Node core

src/generators/node-config-schema/constants.mjs
```diff
@@ -0,0 +1,51 @@
...
+      additionalProperties: false,
+      properties: {},
+      type: 'object',
+    },
+  },
+  type: 'object',
+};
+
+// Schema Definition Map for Data Types
+export const TYPE_DEFINITION_MAP = {
```
flakey5: Once this generator is in use, there should be a comment made in Node core to update this mapping if a new option type is added

flakey5 (approved):
Lgtm, I don't necessarily like using regex here vs what it done in core (using `getEnvOptionsInputType`), but I can't think of anything better unless we want to require the `--expose-internals` flag to run this generator but that's kinda awkward

avivkeller:
We couldn't really use the `--expose-internals` method, even if we wanted to. We are often running `api-docs-tooling` on a different Node.js version than we are generating the documentation for.

src/generators/node-config-schema/index.mjs
```diff
@@ -0,0 +1,142 @@
...
+    // Error handling if either cc or h file is missing
+    if (!ccFile || !hFile) {
+      throw new Error(ERRORS.missingCCandHFiles);
+    }
+
+    // Read the contents of the cc and h files
+    const ccContent = await readFile(ccFile, 'utf-8');
+    const hContent = await readFile(hFile, 'utf-8');
+    const schema = JSON.parse(
+      await readFile(new URL('./schema.json', import.meta.url))
```
ovflowd, line 50: You don't need to do a readFile. You can import the JSON.

src/generators/node-config-schema/index.mjs
```diff
@@ -0,0 +1,142 @@
...
+   * Generates the `node.config.json` schema.
+   * @param {unknown} _ - Unused parameter
+   * @param {Partial<GeneratorOptions>} options - Options containing the input file paths
+   * @throws {Error} If the required files node_options.cc or node_options.h are missing or invalid.
+   */
+  async generate(_, options) {
+    let ccFile, hFile;
+
+    // Ensure input files are provided and capture the paths
+    for (const filePath of options.input) {
```
ovflowd, line 33: Can we use array.find instead?

avivkeller:
@nodejs/config to verify this schema

src/generators/node-config-schema/index.mjs
```diff
@@ -0,0 +1,137 @@
...
+    const { nodeOptions } = schema.properties;
+
+    // Process the cc content and match AddOption calls
+    for (const [, option, config] of ccContent.matchAll(ADD_OPTION_REGEX)) {
+      // If config doesn't include 'kAllowedInEnvvar', skip this option
+      if (!config.includes('kAllowedInEnvvar')) {
+        continue;
+      }
+
+      const headerKey = config.match(OPTION_HEADER_KEY_REGEX)?.[1];
```
ovflowd, line 57:
```suggestion
      const headerKey = config.match(OPTION_HEADER_KEY_REGEX)?.[1];

```

src/generators/node-config-schema/index.mjs
```diff
@@ -0,0 +1,137 @@
...
+    case 'uint64_t':
+    case 'int64_t':
+    case 'HostPort':
+      return { type: 'number' };
+    case 'std::string':
+      return { type: 'string' };
+    case 'bool':
+      return { type: 'boolean' };
+    default:
+      throw new Error(
```
ovflowd, line 133: Would this ever happen?

src/generators/node-config-schema/index.mjs
```diff
@@ -0,0 +1,137 @@
...
+    const ccFile = options.input.find(filePath =>
+      filePath.endsWith('node_options.cc')
+    );
+    const hFile = options.input.find(filePath =>
+      filePath.endsWith('node_options.h')
+    );
+
+    // Error handling if either cc or h file is missing
+    if (!ccFile || !hFile) {
+      throw new Error(ERRORS.missingCCandHFiles);
```
ovflowd, line 41: Do we throw errors on other generators? How does the error look when thrown? As a stack trace?

avivkeller: Yes, we throw errors in other generators, and it looks like a stack trace (See #252)

src/generators/node-config-schema/index.mjs
avivkeller: On the rare case that a new type is added / an invalid type exists, this would throw. If/when this exists in node core, it would throw during a automated update PR, and we'd adjust the which statement accordingly.

https://github.com/nodejs/api-docs-tooling/pull/253#discussion_r2052656901

ovflowd: Noted. How can we catch this? An unit test?

avivkeller: Yea, it'd be nice to add a unit test in node core for this, ideally to run whenever api-docs-tooling is bumped. Once the new tooling is adopted, I'll open a follow-up to generate this file, with the appropriate tests.

marco-ippolito:
Im not sure I understand the goal of this PR, the schema needs to be generated by the node.js version for which the schema is meant

avivkeller:
> Im not sure I understand the goal of this PR, the schema needs to be generated by the node.js version for which the schema is meant

This PR aims to migrate schema generation into the `api-docs-tooling` generator. This allows the schema to be generated using different Node.js versions, similar to how documentation generation is handled.

As demonstrated in this PR, the schema can be generated using a different Node.js version, as long as the files can be parsed.

marco-ippolito (dismissed):
Regexing on a .cc file is not a good idea, I dont see the benefit, there is an internal api designated for this specific purpose

avivkeller:
FWIW https://github.com/nodejs/node/blob/main/test/parallel/test-cli-node-options-docs.js#L20

avivkeller:
@marco-ippolito The internal API requires that the schema to be generated using the same Node.js version it’s intended for.

Ideally, this generator should support generating schemas with previous versions—specifically the LTS versions used in CI.

marco-ippolito:
Whats wrong with the current way of generating the schema?

avivkeller:
> Whats wrong with the current way of generating the schema? It seems to be working fine

There's nothing "wrong" with it, however, the schema must be generated by the same version of Node.js that it is being generated for, which isn't consistent with the rest of `api-doc-tooling`, which intends to build similar things (i.e. Documentation, Man-page) using LTS versions of Node.js, so they can be run in the CI environment on stable versions.

marco-ippolito:
> > Whats wrong with the current way of generating the schema? It seems to be working fine
> 
> There's nothing "wrong" with it, however, the schema must be generated by the same version of Node.js that it is being generated for, which isn't consistent with the rest of `api-doc-tooling`, which intends to build similar things (i.e. Documentation, Man-page) using LTS versions of Node.js, so they can be run in the CI environment on stable versions.

The schema is generated at release time just like api docs. Id rather have this costraint than regex parse a .cc file

ovflowd:
@marco-ippolito, if you still have concerns, you can either block the upstream PR on nodejs/node or block this PR as a TSC member. However, I would urge you to reconsider if possible. :bow:

marco-ippolito:
> Additionally, your block has no authority over this repository.
LOL this is new to me, since when I am not authorized to block a PR as a collaborator @nodejs/tsc ?
 
Since I created the `node.config.json` I think I have the authority to say I don't think regex parsing the node_options.cc file is a good idea since there are some caveats with allowed flags (we dont allow v8 flags and no-ops). The "we have been doing this already" argument doesn't seem right, I created an internal API specifically for this purpose.
I oppose to this specific use case I don't care about the old usage.

marco-ippolito:
@ovflowd can you point to me to the doc saying I cannot block a PR in the doc tooling repo?
Also I noticed you review my block without discussing it first. This feels to me disrespectful and incorrect

RaisinTen:
> if you still have concerns, you can either block the upstream PR on nodejs/node or block this PR as a TSC member

Marco is a TSC member, so based on your words, it's okay for him to block.

ovflowd:
> this is new to me, since when I am not authorized to block a PR as a collaborator @nodejs/tsc ?

It is my understanding that repositories within different working groups and teams operate under their own governance policies. I could be mistaken, but this is my interpretation.

> Since I created the node.config.json I think I have the authority to say I don't think regex parsing the node_options.cc file is a good idea to create the schema since there are some caveats with allowed flags (we dont allow v8 flags and no-ops).

I don't believe the argument of "I created this, so you should do what I say" is a valid one. However, could you elaborate on your point about using regex to parse the `node_options.cc` file to create the schema? I have some concerns regarding this approach, especially considering there are certain caveats with the allowed flags. For instance, we don't permit V8 flags or no-ops.

> The "we have been doing this already" argument doesn't seem right

I'm not saying that either, but it holds its value. We want to make an API-doc-tooling that is as decoupled from the source code as possible.

> Also I noticed you removed my block without discussing it first. This feels to me disrespectful and incorrect

I'm sorry, that was not my intent; You are more than free to raise the block again, there's no need to be disrespectful here :)

marco-ippolito:
> It is my understanding that repositories within different working groups and teams operate under their own governance policies. I could be mistaken, but this is my interpretation.

So this group policy is that a collaborator review can be dismissed without discussion?
That doesnt seem a good policy and its definitely not written anywhere

>  However, could you elaborate on your point about using regex to parse the node_options.cc file to create the schema? Especially "considering there are certain caveats with the allowed flags. For instance, we don't permit V8 flags or no-ops."

You could have asked me to elaborate more but you went ahead and removed my review

I have no interest in interacting with this PR anymore, since its clear my opinion is not respected.

ovflowd:
> So this group policy is that a collaborator review can be dismissed without discussion?

No, I'm not dismissing your discussion, but my understanding is that your block as a core collaborator doesn't hold here (in this repository particularly)... Again I could be wrong. That doesn't mean we will merge the PR without resolving the pending discussions, that's still definitely the goal!

ovflowd:
> You could have asked me to elaborate more but you went ahead and removed my review

I dismissed a :x: marker, a GitHub review; That doesn't mean at all I am dismissing your opinions :bow: 

> I have no interest in interacting with this PR anymore, since its clear my opinion is not respected.

I believe your approach is overly aggressive; this attitude isn't what I expect from a TSC member. We're all aiming for consensus, and I genuinely respect and value your opinions. 🙏

ovflowd:
> > if you still have concerns, you can either block the upstream PR on nodejs/node or block this PR as a TSC member
> 
> Marco is a TSC member, so based on your words, it's okay for him to block this PR.

My understanding is that his block was as a collaborator and not a TSC block, that of course was my understanding of his actions :sweat_smile:

ovflowd:
> > > It is my understanding that repositories within different working groups and teams operate under their own governance policies. I could be mistaken, but this is my interpretation.
> 
> That is true for Working Groups but not teams.

Noted! Apologies, again @marco-ippolito, as my actions seem to have been misplaced. I understand you don't want to interact with this PR anymore, but as we are sunsetting the existing api-doc-tooling, we need some guidance on how to proceed with the node.config generation. Would you be so kind to guide us here?

ShogunPanda:
My 2 cents on the objection dismissal.

First of all, Claudio, let me clarify I totally believe in good faith.
But I think you rushed a little bit (and this is coming from a professional rusher :)). 

Whether Marco was or was not a TSC member was not the point, but rather than I also felt like his opinion was totally ignored with the dismissal. Being a TSC doesn't make his opinion more or less important. It's just about the person.

My (unsolicited) advice for the next time is eventually pointing out in a comment that the objection might not have been binding according to the rules. Then, after thoroughly have gone for consensus seeking as we usually do, eventually override the objection and move on with the PR.

ovflowd (changes requested):
Blocking the PR on behalf of @marco-ippolito as me dismissing their reviews was not an accurate representation of the governance docs and it unnecessarily dismissed Marco's opinions which still hold true.

src/generators/node-config-schema/constants.mjs
```diff
@@ -0,0 +1,15 @@
...
+  missingCCandHFiles:
+    'Both node_options.cc and node_options.h must be provided.',
+  headerTypeNotFound:
+    'A type for "{{headerKey}}" not found in the header file.',
+  missingTypeDefinition: 'No type schema found for "{{type}}".',
+};
+
+// Regex pattern to match calls to the AddOption function.
+export const ADD_OPTION_REGEX =
+  /AddOption[\s\n\r]*\([\s\n\r]*"([^"]+)"(.*?)\);/gs;
```
joyeecheung, line 12: Do I understand correctly that this parses the C++ source code to get the options? If so this feels like a misstep. The current C++ option parser puts an overhead at runtime to add the options. Ideally that should be refactored to add most of the options at compile time and let the parser be statically generated to eliminate the overhead and eliminate more static STL containers. Parsing the source code in an ad-hoc way in another tooling that would break the tests makes it harder to change the parser in the future. It would be better to use the binary to get the options instead of counting on the source code to be a particular shape.

ovflowd: Thanks @joyeecheung for the feedback! I feel the straightforward way is [to simply use the existing code](https://github.com/nodejs/node/blob/2244a0977451c16e7cade9f1aeb46e1c136669bb/tools/doc/generate-json-schema.mjs#L4) incorporated on this repository? Or should the generation of the `node.config.json` schema simply not be handled by the `api-docs-tooling` :thinking: any thoughts or alternatives are appreciated :bow:

ovflowd:
Hey folks :wave: I've chatted with @marco-ippolito and we reached on the following:

1. The `node.config.json` schema should not be part of the API dos tooling, as it is not a "doc" per see but more of a machine-readable JSON schema for places such as IDEs
2. Ill open a PR against nodejs/node to move this generator to the root of `tools`
3. Ill open another PR to move the dest file to be directly within the /dist (release assets) of the output of `node`'s build. (cc @flakey5 do we need to update anything for this file to be uploaded to worker and what not?)
4. @flakey5 we'll might need to do a redirect on the worker for /docs/xxx/node.config.json to /dist/xxx/node.config.json
5. @marco-ippolito or someone else to open a PR on nodejs/node to make the API public instead of internal.

@avivkeller how do you feel with this? Does it sound like a good path to proceed?

src/generators/node-config-schema/constants.mjs
ovflowd: Thanks Joyee, that feels like a good path forward.

avivkeller:
Sounds good to me! If relevant, you may also want to close/update #216 with the additional information

flakey5:
> @flakey5 do we need to update anything for this file to be uploaded to worker and what not

For uploading it to `nodejs/release/vX.X.X/`? I don't think there would be anything except changing the actual output location for the file here https://github.com/nodejs/node/blob/102d8cff66550777bec7b0a750ae2203007f3749/Makefile#L812, definitely could be wrong though

PR #267, opened 2025-05-04
chore(dep): remove useless dep

AugustinMauroy:
## Description

Removing useless dep, we can just use node.

## Validation

```
$npx api-docs-tooling generate  -i ../node/doc/api/*.md -t llms-txt -o out/
(node:15394) ExperimentalWarning: globSync is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Missing 'introduced_in' field in the API doc entry at doc/api/synopsis.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/async_context.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/esm.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/index.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/module.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/packages.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/synopsis.md
Missing llm_description field or paragraph node in the API doc entry at doc/api/typescript.md
```

## Related Issues

No related issues 

### Check List

- [ ] Run `node --run test`
- [ ] Run `node --run lint:fix; node --run format`
- **NA** I've covered new added functionality with unit tests if necessary.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --disable-warning=11982
 
 import process from 'node:process';
 import { Argument, Command, Option } from 'commander';
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "dedent": "^1.5.3",
         "estree-util-visit": "^2.0.0",
         "github-slugger": "^2.0.0",
-        "glob": "^11.0.1",
         "hast-util-to-string": "^3.0.1",
         "hastscript": "^9.0.1",
         "html-minifier-terser": "^7.2.0",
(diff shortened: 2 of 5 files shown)


avivkeller (changes requested):
We tried this before, but agreed not to use glob while it's still experimental

AugustinMauroy:
I don't see what the problem is with using this api, which is experimental but very stable.

ovflowd:
> I don't see what the problem is with using this api, which is experimental but very stable.

how can something experimental be stable. Also, we decided not so because otherwise you cannot run api-docs-tooling on older versions of Node...

AugustinMauroy:
> how can something experimental be stable.

I mean there are not too much bug with it.

> Also, we decided not so because otherwise you cannot run api-docs-tooling on older versions of Node...

Do we want that ? And what should the node supporter versions be?

avivkeller:
The LTS versions are v20 and v22. v20 does not currently have glob.

PR #270, opened 2025-05-12
fix: CryptoKey.algorithm type and other interfaces

flakey5:
Mirrors changes in nodejs/node#58294

### DIFF
--- a/src/utils/parser/constants.mjs
+++ b/src/utils/parser/constants.mjs
@@ -219,7 +219,7 @@ export const DOC_TYPES_MAPPING_NODE_MODULES = {
   RecordableHistogram:
     'perf_hooks.html#class-recordablehistogram-extends-histogram',
 
-  RsaHashedKeyAlgorithm: 'webcrypto.html#class-rsahashedkeygenparams',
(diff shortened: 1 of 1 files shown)


avivkeller (approved):

ovflowd (approved):

panva:
Note the update: https://github.com/nodejs/node/pull/58294/commits

src/utils/parser/constants.mjs
```diff
@@ -214,6 +219,7 @@ export const DOC_TYPES_MAPPING_NODE_MODULES = {
   RecordableHistogram:
     'perf_hooks.html#class-recordablehistogram-extends-histogram',
 
+  RsaHashedKeyAlgorithm: 'webcrypto.html#class-rsahashedkeygenparams',
```
panva, line 222:
```suggestion
  RsaHashedKeyAlgorithm: 'webcrypto.html#class-rsahashedkeyalgorithm',
```

avivkeller:
Adding the https://github.com/nodejs/api-docs-tooling/labels/blocked label since we should wait for this to land upstream before landing it here, right?

bjohansebas (approved):

Issue #236, opened 2025-03-24
How are we going to import `@node-core/ui-components`?


avivkeller:
npm does not support importing packages from subdirectories of git repositories.

According to https://github.com/npm/npm/issues/2974#issuecomment-169451651:  
> This isn't something that the npm CLI team is interested in adding to the tool.  

There are third-party solutions like [GitPkg](https://gitpkg.vercel.app/about) that allow this functionality in npm, but I'm not sure we should rely on a third-party tool to handle imports–lest something bad happen. 

**However**, both [Yarn](https://github.com/yarnpkg/yarn/issues/4725#issuecomment-753416391) (although, yarn requires the website to also use yarn) and [pnpm](https://github.com/pnpm/pnpm/pull/7487) do support this, so we *could* convert this repository to use one of them instead.

This would also require any project that runs `api-docs-tooling` use the same package manager (i.e. `pnpm dlx` instead of `npx` for execution).  

Reference: https://github.com/nodejs/nodejs.org/pull/7401#issuecomment-2748180596

AugustinMauroy:
Why not publish ui-components ?

avivkeller:
That's an option, but I feel like we are going to be constantly updating that package, so it might be easier to keep it as a repository.

avivkeller:
No, we'd just change the npm installation step to a pnpm installation step, but let's cross that bridge if we get to it.

avivkeller:
We wouldn't need to depend on it, we would just have to assume the user has it installed, which isn't ideal, but every one of our options has a drawback.

- Switching package managers has the drawbacks of requiring Node.js doc-builders and linters to have it installed

- Publishing to npm has the drawback of needing to setup a publishing automation, and not control versions by commit

- Using a third party tool has the drawback of being a third party tool

ovflowd:
> > It can be as an GitHub artifact.
> 
> That could work. Once the `ui-components` library lands, I'll make a CI/CD pipeline for uploading it to the GitHub packages registry.
> 
> I'm gonna leave this issue open until then, just in case anything changes.

Are you merging the PR today?

avivkeller:
Yes, I don't think anyone else is going to object in the next hour, so I'm merge it now, but I'm going to add a workflow in a second PR, in case it needs a revert, we don't need to revert the entire UI package

avivkeller:
[...]

PR #273, opened 2025-05-17
[WIP] mdx generator

avivkeller:
Right now, this PR only contains the basic code to convert the markdown nodes into MDX nodes, so it's an extremely oh-so-very early draft.

The additional dependencies are purely for testing the output of this generator, which you can do by plugging it into https://github.com/avivkeller/nodejs-docs-redesign-reference as `src/content.mdx`.


avivkeller:
Current output:
image

package.json
```diff
@@ -47,6 +47,8 @@
     "hast-util-to-string": "^3.0.1",
     "hastscript": "^9.0.1",
     "html-minifier-terser": "^7.2.0",
+    "mdast-util-to-markdown": "^2.1.2",
+    "mdast-util-mdx-jsx": "^3.2.0",
```
avivkeller, line 51: These are here for testing the output, and will be removed before this is ready for review

AugustinMauroy (commented):
Interesting basic, don't miss to add some test

avivkeller:
I'm working on skipping the MDX and going to straight to JSX, so I'll add lots of tests when that's started

avivkeller:
This is a VERY BIG work in progress.

PNPM because of #236 (which will ideally be a different PR)
Webpack because the imports from `@node-core` aren't transpiled, although I hope to remove webpack before this is ready for review.

src/generators/jsx/constants.mjs
```diff
@@ -0,0 +1,6 @@
+export const STABILITY_LEVELS = [
```
AugustinMauroy, line 1: Something like that:
```suggestion
/* 
 * This allow use to link doc stability level to alertBox props
 */
export const STABILITY_LEVELS = [
```

avivkeller:
This uses pnpm for now, see #236 for reference

src/generators/index.mjs
```diff
@@ -23,6 +24,7 @@ export const publicGenerators = {
   'api-links': apiLinks,
   'orama-db': oramaDb,
   'llms-txt': llmsTxt,
+  jsx,
```
AugustinMauroy, line 27:
```suggestion
  'jsx': jsx,
```

avivkeller: What's the difference?

AugustinMauroy: same but rest use so why not use same

avivkeller: The rest use different keys than the variable names, so we wouldn't be able to use this syntax. 

I don't think we need to change it.

src/utils/generators.mjs
```diff
@@ -53,3 +53,43 @@ export const coerceSemVer = version => {
...
+  if (!introduced) return releases;
+  const coercedMajor = major(coerceSemVer(introduced));
+  // All Node.js versions that support the current API; If there's no "introduced_at" field,
+  // we simply show all versions, as we cannot pinpoint the exact version
+  return releases.filter(release => release.version.major >= coercedMajor);
+};
+
+/**
+ * Maps `updates` into `changes` format, merges them and sorts them by version
+ * ç
```
canerakdas, line 74:
```suggestion
 *
```
Looks like I pressed a random key 🙈

package.json
```diff
@@ -15,7 +15,8 @@
     "test:watch": "node --test --watch",
     "prepare": "husky",
     "run": "node bin/cli.mjs",
-    "watch": "node --watch bin/cli.mjs"
+    "watch": "node --watch bin/cli.mjs",
+    "postinstall": "git-deps install"
```
anonrig, line 19: If this is the intended behavior and the solution, we should move this dependency to node.js organization, before merging this pull-request. It will run on every postinstall, and it's huge security concern.

anonrig (dismissed):

package.json
avivkeller: We are still looking for solutions, see #236

ovflowd: Agreed with @anonrig although I'd say this is a draft, but we def... need to find a perma solution.

ovflowd: We should at the very least move this dep to @pkgjs. @avivkeller feel free to open an issue on /admin to transfer this + ownership.

avivkeller: Shouldn't we first decide whether this is a perma solution? I'd argue that we shouldn't maintain a package that we aren't going to use-but I'm happy to transfer it to the org regardless

ovflowd: Sure, but this PR cannot land till we decide that, or at least cannot land whilst git-deps is a dependency here :sweat_smile:

src/generators/jsx/utils/buildBarProps.mjs
```diff
@@ -0,0 +1,51 @@
...
+    };
+  });
+
+/**
+ * Builds metadata for the sidebar and meta bar
+ *
+ * @param {ApiDocMetadataEntry} head - Main API metadata entry
+ * @param {Array<ApiDocMetadataEntry>} entries - All API metadata entries
+ */
+export function buildMetaBarProps(head, entries) {
```
ovflowd, line 30: Maybe explain a bit more why we need to append all this text

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Transforms a stability node into an AlertBox JSX element
+ *
+ * @param {import('mdast').Blockquote} node - The stability node to transform
+ * @param {number} index - The index of the node in its parent's children array
+ * @param {import('unist').Parent} parent - The parent node containing the stability node
+ * @returns {[typeof SKIP]} Visitor instruction to skip the node
+ */
+function transformStabilityNode(node, index, parent) {
```
ovflowd, line 26: nit: Maybe add a blank line between 27-28 and 32-33?

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,199 @@
...
+  if (entry.changes?.length) {
+    const explicitChanges = entry.changes.map(change => ({
+      versions: enforceArray(change.version),
+      label: change.description,
+      url: change['pr-url'],
+    }));
+    changeEntries.push(...explicitChanges);
+  }
+
+  if (!changeEntries.length) return null;
```
ovflowd, line 64: {} brackets

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/**
+ * Creates a source link element if a source link is available
+ *
+ * @param {string|undefined} sourceLink - The source link path
+ * @returns {import('hastscript').Element|null} The source link element or null if no source link
+ */
+function createSourceLink(sourceLink) {
+  if (!sourceLink) return null;
```
ovflowd, line 79: {}

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * @param {ApiDocMetadataEntry} entry - The metadata entry containing change information
+ * @returns {import('unist').Node|null} JSX element representing change history or null if no changes
+ */
+function createChangeElement(entry) {
+  // Collect changes from version fields (added, deprecated, etc.)
+  const changeEntries = Object.entries(CHANGE_TYPES)
+    .filter(([field]) => entry[field])
+    .map(([field, label]) => {
+      const versions = enforceArray(entry[field]);
```
ovflowd, line 47: Maybe call another map to apply the enforceArray, so that the last map can just return `({ versions, label: ... })`

package.json
avivkeller: > Sure, but this PR cannot land till we decide that

Per the vote in https://github.com/nodejs/api-docs-tooling/issues/236, it looks like https://github.com/nodejs/nodejs.org/pull/7776 will resolve that, and once they are published, I'll switch it. `git-deps` allows me to test this PR until then.

src/generators/jsx/utils/ast.mjs
```diff
@@ -0,0 +1,74 @@
...
+ */
+export const createJSXElement = (
+  name,
+  { inline = true, children = [], ...attributes } = {}
+) => {
+  // Process children: convert string to text node or use array as is
+  const processedChildren =
+    typeof children === 'string'
+      ? [createTree('text', { value: children })]
+      : (children ?? []);
```
ovflowd, line 25: What types children can be here?

src/generators/jsx/utils/ast.mjs
```diff
@@ -0,0 +1,74 @@
...
+            ],
+            sourceType: 'module',
+          },
+        },
+      }),
+    });
+  }
+
+  // For primitives, use simple string conversion
+  return createTree('mdxJsxAttribute', {
```
ovflowd, line 70: This can be inline no? (the object)

src/generators/jsx/utils/buildBarProps.mjs
avivkeller: Like what, does `// Extract text content for reading time calculation` not sum it up? `readingTime` expects text, not AST.

src/generators/jsx/utils/ast.mjs
avivkeller: Anything that's a valid Markdown AST

src/generators/jsx/utils/ast.mjs
```diff
@@ -0,0 +1,74 @@
...
+
+import { u as createTree } from 'unist-builder';
+import { valueToEstree } from 'estree-util-value-to-estree';
+
+/**
+ * Creates an MDX JSX element with support for complex attribute values.
+ *
+ * @param {string} name - The name of the JSX element
+ * @param {{
+ * inline?: boolean,
```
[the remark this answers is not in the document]
ovflowd: Right, but I believe extra level, makes it easier to read:

```
/**
 * Creates an MDX JSX element with support for complex attribute values.
 *
 * @param {string} name - The name of the JSX element
 * @param {{
 *   inline?: boolean,
 *   children?: string | import('unist').Node[],
 *   [key: string]: any
 * }} [options={}] - Options including type, children, and JSX attributes
 * @returns {import('unist').Node} The created MDX JSX element node
 */
```

src/generators/jsx/utils/buildBarProps.mjs
```diff
@@ -0,0 +1,51 @@
...
+    headings: entries
+      .filter(entry => entry.heading?.data?.name)
+      .map(entry => ({
+        depth: entry.heading.depth,
+        value: entry.heading.data.name,
+      })),
+    addedIn: head.introduced_in || head.added_in || '',
+    readingTime: readingTime(textContent).text,
+    viewAs: [['JSON', `${head.api}.json`]],
+    editThisPage: `${DOC_API_BLOB_EDIT_BASE_URL}${head.api}.md`,
```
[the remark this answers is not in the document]
ovflowd: Might make sense to then make this as a property on the metadata itself

src/generators/jsx/utils/buildBarProps.mjs
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Builds metadata for the sidebar and meta bar
+ *
+ * @param {ApiDocMetadataEntry} head - Main API metadata entry
+ * @param {Array<ApiDocMetadataEntry>} entries - All API metadata entries
+ */
+export const buildMetaBarProps = (head, entries) => {
+  // Extract text content for reading time calculation
+  let textContent = '';
```
ovflowd, line 32: Let's use Array.reduce here instead of an outside let that can be reassigned anytime and a forEach.

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+ * Creates the overall content structure with processed entries
+ *
+ * @param {Array<ApiDocMetadataEntry>} entries - API documentation metadata entries
+ * @param {Object} sideBarProps - Props for the sidebar component
+ * @param {Object} metaBarProps - Props for the meta bar component
+ * @returns {import('unist').Node} The root node of the content tree
+ */
+const createContentStructure = (entries, sideBarProps, metaBarProps) => {
+  return createTree('root', [
+    createJSXElement('NavBar'),
```
ovflowd, line 166: Since all these are React Components, would make sense to _reference_ these as constants such as:

```
import { NavBar } from '@node-core/ui....'

...

createJSXElement(NavBar.displayName), // or something similar
```

This ensures that we use their real names; And then wherever we expose them (be it next.mdx.mjs and next.client.mdx.mjs) we also use the same constants.

This ensures that this won't randomly break if we chan the declared Components within MDX or somewhere else. Apply this to all components you're invoking within this new generator.

src/generators/jsx/utils/transformer.mjs
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {Object} tree - The abstract syntax tree to transform
+ * @returns {Object} The transformed tree (modified in place)
+ */
+export default () => tree => {
+  visit(tree, ['raw', 'element'], node => {
+    // TODO(@avivkeller): Our parsers shouldn't return raw nodes
+    // When they mistake "<Type>" for an HTML node, rather than
+    // the string type that it is.
+    if (node.type === 'raw') {
```
ovflowd, line 15: I'd rather have a switch statement here :P

src/generators/jsx/utils/buildBarProps.mjs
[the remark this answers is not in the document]
avivkeller: I don't think it's needed. It's really just `{constant}{metadata.api}.md`, so it can very easily be built, however, if you insist, maybe that can be a follow up.

src/generators/jsx/utils/ast.mjs
```diff
@@ -0,0 +1,71 @@
...
+        data: {
+          estree: {
+            type: 'Program',
+            body: [
+              {
+                type: 'ExpressionStatement',
+                expression: valueToEstree(value),
+              },
+            ],
+            sourceType: 'module',
```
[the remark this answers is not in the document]
avivkeller: I was following [the example code](https://github.com/syntax-tree/mdast-util-mdx-jsx), but we are only using this for objects, so I don't think it matters, let me test that.

src/generators/jsx/utils/transformer.mjs
avivkeller: > I'd rather have a switch statement here :P

There's only one clause, so I think we should leave it as is.

> Or simply have `text` within `TYPE_TRANSFORMS`

The transforms we are apply are for `.tagName`, not `.type`

src/generators/jsx/constants.mjs
```diff
@@ -0,0 +1,51 @@
...
+  method: { symbol: 'M', color: 'red' },
+  property: { symbol: 'P', color: 'red' },
+  class: { symbol: 'C', color: 'red' },
+  module: { symbol: 'M', color: 'red' },
+  classMethod: { symbol: 'S', color: 'red' },
+  ctor: { symbol: 'C', color: 'red' },
+};
+
+// Maps API lifecycle change type identifiers to their human-readable labels.
+export const CHANGE_TYPES = {
```
ovflowd, line 27: IMO, we should import @node-core/i18n and have these as translation strings. This can be done in a follow-up PR. To be clear, @nodejs/web-infra we shouldn't translate the API docs, but at the very least, navigational and non-main-content markers can be translated.

src/generators/jsx/utils/ast.mjs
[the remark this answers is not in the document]
ovflowd: Noted, then because of the lint rule _this got changed_ because when I started this repo I 100% did this sort of indentation xD

I'm fine leaving as it is (feel free to resolve this thread)... but, IMO this makes readability better.

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+ * @param {string|undefined} sourceLink - The source link path
+ * @returns {import('hastscript').Element|null} The source link element or null if no source link
+ */
+const createSourceLink = sourceLink => {
+  if (!sourceLink) {
+    return null;
+  }
+
+  return createElement('span', [
+    'Source Code: ',
```
ovflowd, line 87: Another constant (that should also be translateable)

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Transforms a stability node into an AlertBox JSX element
+ *
+ * @param {import('mdast').Blockquote} node - The stability node to transform
+ * @param {number} index - The index of the node in its parent's children array
+ * @param {import('unist').Parent} parent - The parent node containing the stability node
+ * @returns {[typeof SKIP]} Visitor instruction to skip the node
+ */
+const transformStabilityNode = ({ data }, index, parent) => {
+  parent.children[index] = createJSXElement('AlertBox', {
```
ovflowd, line 27: Another JSX component

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+    }));
+
+  // Add explicit changes if they exist
+  if (entry.changes?.length) {
+    const explicitChanges = entry.changes.map(change => ({
+      versions: enforceArray(change.version),
+      label: change.description,
+      url: change['pr-url'],
+    }));
+    changeEntries.push(...explicitChanges);
```
ovflowd, line 62:
```suggestion

    changeEntries.push(...explicitChanges);
```

nit: readability

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+const transformHeadingNode = (entry, node, index, parent) => {
+  const { data, children } = node;
+  const headerChildren = [
+    createElement(`h${data.depth + 1}`, [
+      createElement(`a.mark#${data.slug}`, { href: `#${data.slug}` }, children),
+    ]),
+  ];
+
+  // Add type icon if available
+  const iconSymbol = ICON_SYMBOL_MAP[data.type];
```
ovflowd, line 114: I don't think you need to access a constant here. Also:

```js
if (data.type in ICON_SYMBOL_MAP) {
```

src/generators/jsx/utils/buildContent.mjs
```diff
@@ -0,0 +1,201 @@
...
+  if (iconSymbol) {
+    headerChildren.unshift(createJSXElement('CircularIcon', iconSymbol));
```
ovflowd, line 116:
```suggestion
    headerChildren.unshift(createJSXElement('CircularIcon', ICON_SYMBOL_MAP[data.type]));
```

src/generators/jsx/utils/buildContent.mjs
ovflowd: So you import the constants, and then once they get IMPL you can update the constants file.

BTW, open an issue for this if you want to do it as a follow-up

src/generators/jsx/utils/transformer.mjs
ovflowd: I got that, but you are doing a check both for .type and .tagName; So you can have a map for .tagName (TAG_TRANSFORMS) and one for .type (TYPE_TRANSFORMS) in a way that this can be simplified to:

```jsx
node.type = node.type in TYPE_TRANSFORMS ? TYPE_TRANSFORMS[node.type] : node.type;
node.tagName = node.tagName in TAG_TRANSFORMS ? TAG_TRANFORMS[node.tagName] : node.tagName;
```

ovflowd: So pretty much your code below can be simplified to the code above and reduce mental strain

src/utils/remark.mjs
```diff
@@ -37,3 +42,30 @@ export const getRemarkRehype = () =>
...
+    .use(remarkParse)
+    // We make Rehype ignore existing HTML nodes, and JSX nodes
+    // as these are nodes we manually created during the generation process
+    // We also allow dangerous HTML to be passed through, since we have HTML within our Markdown
+    // and we trust the sources of the Markdown files
+    .use(remarkRehype, {
+      allowDangerousHtml: true,
+      passThrough: [
+        'element',
+        'mdxFlowExpression',
```
ovflowd, line 61: Use the conmstants here

src/utils/generators.mjs
```diff
@@ -53,3 +53,43 @@ export const coerceSemVer = version => {
 
   return coercedVersion;
 };
+
+/**
+ * Gets compatible versions for an entry
+ *
+ * @param {string | import('semver').SemVer} introduced
+ * @param {Array<Release>} releases
```
canerakdas, line 61: It may be useful to add the `Release` object used here as a type, since it is not currently defined; everything under this object currently appears as `any` 🙇

src/utils/generators.mjs
```diff
@@ -53,3 +53,43 @@ export const coerceSemVer = version => {
...
+
+    return compare(coerceSemVer(aVersion), coerceSemVer(bVersion));
+  });
+};
+
+/**
+ * Converts a value to an array.
+ * @template T
+ * @param {T | T[]} val - The value to convert.
+ * @returns {T[]} The value as an array.
```
canerakdas, line 93:
```suggestion
 * @param {T | Array<T>} val - The value to convert.
 * @returns {Array<T>} The value as an array.
```

src/generators/jsx/utils/ast.mjs
```diff
@@ -0,0 +1,77 @@
...
+        },
+      }),
+    });
+  }
+
+  // For primitives, use simple string conversion.
+  // If undefined, pass nothing.
+  return createTree(AST_NODES.MDX.JSX_ATTRIBUTE, {
+    name,
+    value: value === undefined ? value : String(value),
```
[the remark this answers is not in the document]
avivkeller: That's intentional, in case we need an explicitly null value.

Actually, I should verify Recma can handle that scenario.

package.json
avivkeller: I'm resolving this disucssion, and adding the `blocked` label until the upstream PR lands.

src/utils/remark.mjs
```diff
@@ -37,3 +42,30 @@ export const getRemarkRehype = () =>
...
+ */
+export const getRemarkRecma = () =>
+  unified()
+    .use(remarkParse)
+    // We make Rehype ignore existing HTML nodes, and JSX nodes
+    // as these are nodes we manually created during the generation process
+    // We also allow dangerous HTML to be passed through, since we have HTML within our Markdown
+    // and we trust the sources of the Markdown files
+    .use(remarkRehype, {
+      allowDangerousHtml: true,
```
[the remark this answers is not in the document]
avivkeller:
```
// We also allow dangerous HTML to be passed through, since we have HTML within our Markdown
// and we trust the sources of the Markdown files
```
Check 55-56

package.json
```diff
@@ -36,17 +36,22 @@
   "dependencies": {
     "@actions/core": "^1.11.1",
     "@clack/prompts": "^0.10.1",
+    "@node-core/rehype-shiki": "^1.0.1-1815fa769361b836fa52cfab9c5bd4991f571c95",
```
ovflowd, line 39: Not importing any element from ui-components?

avivkeller: > Or is this jsx generator just to generate new jsx nodes and then another generator to transform to html? (jsx layouts + all elements + what not)?

Bingo. Perhaps a rename to jsx-ast?

avivkeller: This generator generates the nodes, and the react-web one takes those nodes, compiles it with the components using ESBuild, and outputs HTML

flakey5 (approved):

AugustinMauroy:
maybe add little bit more test

package.json
avivkeller: ESBuild was mentioned in #7, but any compiler works. That's a discussion for that issue.

src/generators/jsx-ast/utils/buildBarProps.mjs
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Builds metadata for the sidebar and meta bar
+ *
+ * @param {ApiDocMetadataEntry} head - Main API metadata entry
+ * @param {Array<ApiDocMetadataEntry>} entries - All API metadata entries
+ */
+export const buildMetaBarProps = (head, entries) => {
+  // Extract text content for reading time calculation
+  const textContent = entries.reduce((acc, entry) => {
```
ovflowd, line 32: Oh, I didn't know the resulting reduce would look like this, this can lead to asynchronous issues. You can use a method visitAll to get all these nodes in one array if all you want is to get a value. cc @avivkeller urgent.

avivkeller: If it's urgent, can you open a fix?

ovflowd: Can wait

ovflowd: You are calling a callback inside another callback, these two are not synchronous. entries.reduce will not wait for visit() if visit() is async. Might not be the case here, but still, rather do a visitAll() and then grab all the values.

avivkeller: `visit` isn't async (and `visitAll` isn't a function)

PR #274, opened 2025-05-17
chore(lint): use `console[xyz]` over `process`

avivkeller:
If all of these are sent to `stdout` and the user pipes the output using `>`, the lint errors will be included. Instead, the outputs should be directed appropriately based on the context, right?

### DIFF
--- a/src/linter/reporters/console.mjs
+++ b/src/linter/reporters/console.mjs
@@ -21,10 +21,10 @@ export default issue => {
     ? ` (${issue.location.position.start.line}:${issue.location.position.end.line})`
     : '';
 
-  process.stdout.write(
+  console[issue.level](
     styleText(
       levelToColorMap[issue.level],
       `${issue.message} at ${issue.location.path}${position}`
-    ) + '\n'
+    )
   );
 };
--- a/src/linter/tests/reporters/console.test.mjs
+++ b/src/linter/tests/reporters/console.test.mjs
@@ -1,26 +1,36 @@
 import { describe, it } from 'node:test';
-import console from '../../reporters/console.mjs';
 import assert from 'node:assert';
+import reporter from '../../reporters/console.mjs';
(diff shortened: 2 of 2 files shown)


src/linter/reporters/console.mjs
```diff
@@ -21,10 +21,10 @@ export default issue => {
     ? ` (${issue.location.position.start.line}:${issue.location.position.end.line})`
     : '';
 
-  process.stdout.write(
+  console[issue.level](
```
flakey5, line 24: Can we make sure this function exists before calling it?

flakey5 (changes requested):

src/linter/reporters/console.mjs
avivkeller: We know the function exists, the level is either "info", "warn", or "error".

src/linter/tests/reporters/console.test.mjs
```diff
@@ -1,26 +1,36 @@
...
+const testCases = [
+  {
+    issue: infoIssue,
+    method: 'info',
+    expected: '\x1B[90mThis is a INFO issue at doc/api/test.md\x1B[39m',
+  },
+  {
+    issue: warnIssue,
+    method: 'warn',
+    expected: '\x1B[33mThis is a WARN issue at doc/api/test.md (1:1)\x1B[39m',
```
AugustinMauroy, line 15: Can we use snapshot ?

AugustinMauroy (approved):
LGTM just maybe on improvement

araujogui (approved):

src/linter/tests/reporters/console.test.mjs
avivkeller: I don't think we use snapshot elsewhere, so maybe you can open a follow up to change them all?

flakey5 (approved):

PR #311, opened 2025-06-05
chore(lint): missing-metadata

avivkeller:
We can make linting faster if we merge the similar llm description and introduced in check into a single pass.


araujogui (commented):
Great PR. The only downside I see is that users can't disable only the introduced-in or llm-description rules.

src/linter/rules/missing-metadata.mjs
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @param {import('../types.d.ts').LintContext} context
+ * @returns {void}
+ */
+export const missingMetadata = context => {
+  const foundMetadata = new Set();
+  let hasParagraph = false;
+
+  findTopLevelEntry(context.tree, node => {
+    if (node.type === 'html') {
```
ovflowd, line 51: You know how I feel with nested ifs and nested for loops within nested ifs 🙈

src/linter/rules/missing-metadata.mjs
```diff
@@ -0,0 +1,77 @@
...
+      for (const check of METADATA_CHECKS) {
+        if (check.regex?.test(node.value)) {
+          foundMetadata.add(check.name);
+        }
+      }
+    } else if (node.type === 'paragraph') {
+      hasParagraph = true;
+    }
+    return false; // Continue searching
```
ovflowd, line 60: Can this create a situation of infinite loop?

avivkeller: No, `findTopLevelEntry` will stop after it searches all the top level entries

ovflowd (approved):
LGTM

PR #316, opened 2025-06-22
feat(index): support passing an index file

### LINKED ISSUE #314: Improve single-file generation
_See comment posted by @aduh95 in https://github.com/nodejs/node/pull/57343#discussion_r2155232448_

We don't really need to completely generate each file every time, so what if we added a `--select <glob>`, which allows the user to include the entire context (via `--input`), but only _select_ specific files for actual generation.

Alternatively,

- `--only-regenerate-changed`, which compares the timestamps of the input to the output
- `--context`, which explicitly has files to be context, but not generated

avivkeller:
Fixes #314

CC @flakey5


src/utils/parser.mjs
```diff
@@ -0,0 +1,15 @@
...
+
+import { readFile } from 'node:fs/promises';
+
+/**
+ * Loads content from a URL or file path
+ * @param {string|URL} url The URL or file path to load
+ * @returns {Promise<string>} The content as a string
+ */
+export const loadFromURL = async url => {
+  const parsedUrl = url instanceof URL ? url : URL.parse(url);
```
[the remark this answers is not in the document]
avivkeller: `URL.parse` has been available in Node.js since v22, https://developer.mozilla.org/en-US/docs/Web/API/URL/parse_static#browser_compatibility

avivkeller:
Ignore the codecov failure. It's because we don't have any tests covering our CLI.

AugustinMauroy (approved):
LGMT !

flakey5 (approved):
Thanks!

avivkeller:
Requesting fast track as to not block the migration

package.json
```diff
@@ -10,11 +10,11 @@
...
     "format:check": "prettier --check .",
-    "test": "node --test",
+    "test": "node --test --experimental-test-module-mocks",
     "test:coverage": "c8 npm test",
-    "test:ci": "c8 --reporter=lcov node --test --test-reporter=@reporters/github --test-reporter-destination=stdout --test-reporter=junit --test-reporter-destination=junit.xml --test-reporter=spec --test-reporter-destination=stdout",
-    "test:update-snapshots": "node --test --test-update-snapshots",
-    "test:watch": "node --test --watch",
+    "test:ci": "c8 --reporter=lcov node --test --experimental-test-module-mocks --test-reporter=@reporters/github --test-reporter-destination=stdout --test-reporter=junit --test-reporter-destination=junit.xml --test-reporter=spec --test-reporter-destination=stdout",
+    "test:update-snapshots": "node --test --experimental-test-module-mocks --test-update-snapshots",
+    "test:watch": "node --run test --watch",
```
ovflowd, line 17:
```suggestion
    "test:watch": "node --run test -- --watch",
```

ovflowd: Otherwise you're running --watch against the node --run command which will give an error asking for --watch-path

src/parsers/markdown.mjs
```diff
@@ -58,4 +70,35 @@ const createParser = linter => {
...
+/**
+ * Retrieves all Node.js major versions from the provided CHANGELOG.md file
+ * and returns an array of objects containing the version and LTS status.
+ * @param {string|URL} path Path to changelog
+ * @returns {Promise<Array<ApiDocReleaseEntry>>}
+ */
+export const parseChangelog = async path => {
+  const changelog = await loadFromURL(path);
+
+  const nodeMajors = Array.from(changelog.matchAll(NODE_VERSIONS_REGEX));
```
ovflowd, line 82: This requires a changelog file to be passed no? Wouldn't this require the changelog argument to be mandatory?

ovflowd (commented):
Overall LGTM, but left a few comments. Will approve after having them resolved.

package.json
avivkeller: Ahh, my intent was to pass it as an argument to run, got it, I'll adjust accordingly

src/parsers/markdown.mjs
avivkeller: `changelog` is defined on line 80

src/releases.mjs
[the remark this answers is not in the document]
avivkeller: Since it shares a lot of the logic with the index parsing, so I made a central `util/` file, and at that point, this file, and the (added) file for index parsing both only contained a single function, so I put them both in the `parser.mjs` file.

avivkeller:
Just waiting for https://github.com/nodejs/api-docs-tooling/pull/318 to land for a green CI, then we should be good to land.

ovflowd (approved):

avivkeller:
@flakey5 Updating the migration to d20f49e8e19a1889c8bb2b2c4c95bec800c4a641 has the updated code.

PR #322, opened 2025-06-25
fix(legacy-json): verify list type

### LINKED ISSUE #321: Non-option lists are misread as options in `parseList`
Take a look at this snippet from https://github.com/nodejs/node/blob/main/doc/api/test.md:
```md
* `count` {number} The number of assertions and subtests that are expected to run.
* `options` {Object} Additional options for the plan.
  * `wait` {boolean|number} The wait time for the plan:
    * If `true`, the plan waits indefinitely for all assertions and subtests to run.
    * If `false`, the plan performs an immediate check after the test function completes,
      without waiting for any pending assertions or subtests.
      Any assertions or subtests that complete after this check will not be counted towards the plan.
    * If a number, it specifies the maximum wait time in milliseconds
      before timing out while waiting for expected assertions and subtests to be matched.
      If the timeout is reached, the test will fail.
      **Default:** `false`.
```

In both the `nodejs/node` generator, and the `legacy-json` generator, the following is parsed to[^1]:
```json
{
  "textRaw": "`options` {Object} Additional options for the plan.",
  "name": "options",
  "type": "Object",
  "desc": "Additional options for the plan.",
  "options": [
    {
      "textRaw": "`wait` {boolean|number} The wait time for the plan:",
      "name": "wait",
      "type": "boolean|number",
      "desc": "The wait time for the plan:",
      "options": [
        {
          "textRaw": "If `true`, the plan waits indefinitely for all assertions and subtests to run.",
          "name": "If",
          "desc": "`true`, the plan waits indefinitely for all assertions and subtests to run."
        },
        {
          "textRaw": "If `false`, the plan performs an immediate check after the test function completes, without waiting for any pending assertions or subtests. Any assertions or subtests that complete after this check will not be counted towards the plan.",
          "name": "If",
          "desc": "`false`, the plan performs an immediate check after the test function completes, without waiting for any pending assertions or subtests. Any assertions or subtests that complete after this check will not be counted towards the plan."
        },
        {
          "textRaw": "If a number, it specifies the maximum wait time in milliseconds before timing out while waiting for expected assertions and subtests to be matched. If the timeout is reached, the test will fail. **Default:** `false`.",
          "name": "If",
          "default": "`false`",
          "desc": "a number, it specifies the maximum wait time in milliseconds before timing out while waiting for expected assertions and subtests to be matched. If the timeout is reached, the test will fail."
        }
      ]
    }
  ]
}
```

As you can see, the `If ...` lines are incorrectly treated as sub-options.

[^1]: With whatever little differences between the generators

avivkeller:
Fixes #321 by verifying the list meets the following structure before parsing it as a list of options:

1. Must start with a `code`
2. Followed by a space
3. Followed by a link whose first character is `<`

### DIFF
--- a/src/generators/legacy-json/utils/__tests__/parseList.test.mjs
+++ b/src/generators/legacy-json/utils/__tests__/parseList.test.mjs
@@ -123,4 +123,50 @@ describe('parseList', () => {
     parseList(section, nodes);
     assert.ok(Array.isArray(section.params));
   });
+
+  it('processes recursive lists', () => {
+    const section = { type: 'event' };
+    const nodes = [
+      {
+        type: 'list',
+        children: [
+          {
+            children: [
+              {
+                type: 'paragraph',
+                children: [
+                  { type: 'text', value: 'param1 {string} first parameter' },
+                ],
+              },
+              // This is a nested typed list
+              {
+                type: 'list',
+                children: [
+                  {
+                    children: [
+                      {
+                        type: 'paragraph',
+                        children: [
+                          { type: 'inlineCode', value: 'option' }, // inline code
+                          { type: 'text', value: ' ' }, // space
+                          {
+                            type: 'link',
+                            children: [{ type: 'text', value: '<boolean>' }], // link with < value
+                          },
+                          { type: 'text', value: ' option description' },
+                        ],
+                      },
+                    ],
+                  },
+                ],
+              },
+            ],
+          },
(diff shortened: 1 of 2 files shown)


AugustinMauroy (approved):

avivkeller:
Feel free to approve, I just need to add tests, and fix one thing, which I’ll do tomorrow.

PR #323, opened 2025-06-26
feat(perf): put css/js in external files for legacy-html

avivkeller:
By not having terser `minifyCSS` or `minifyJS`, we save ~7 seconds on generation. See also #320. After this, Shiki is the hottest code path, followed by `parseApiDoc`.


src/generators/legacy-html/assets/api.js
avivkeller, line 1: The re-write of this file makes it so the js in `template.html` isn't needed at all.

AugustinMauroy:
Hum sad 😔

ovflowd (approved):
Sugar

PR #347, opened 2025-07-09
fix(legacy-json): more robust signature parsing

avivkeller:
Previously, the signature parsing would parse trailing brackets as if they affect the _current_ parameter, rather than the _next_ parameter.

Because of this, `mand[, opt]` would be seen as an optional param named "mand", and a mandatory parameter "opt". Now, trailing brackets correctly apply to the _next_ parameter.


ovflowd (approved):

avivkeller:
@ovflowd I don't have my GPG key (`FB5D4500627F7285`) on this machine 😅, can we disable that rule? We don't have it on `nodejs/nodejs.org`, right?

If not, I understand, and I can generate a temporary GPG key to sign my commits with, just let me know.

PR #362, opened 2025-07-19
fix(win): use windows-safe paths

avivkeller:
Currently, `doc-kit` does not work on Windows due to a Rolldown issue, and a paths issue.

This PR fixes the paths issue by using Windows-safe paths.

Additionally,
- When not minified, there is a chance that the code will end with an `//#endregion` comment. We need to add a newline before the `return`
- We set the variable `code =` to to dehydration, not our generated variable. TBH, I'm not sure how this worked at all previously...

@TheAlexLichter There appears to be a bug when building this on Windows:
It appears to import everything twice. Once using a Unix path, and once using a Windows path.

### DIFF
--- a/src/generators/web/constants.mjs
+++ b/src/generators/web/constants.mjs
@@ -1,3 +1,17 @@
+import { parse, relative, sep, dirname } from 'node:path';
+import { resolve } from 'node:path/posix';
+import { fileURLToPath } from 'node:url';
+
+// Convert the current module's URL to a filesystem path,
+// then calculate the relative path from the system root directory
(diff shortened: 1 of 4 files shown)


AugustinMauroy (approved):

avivkeller:
@TheAlexLicher Per the Rolldown trace logs, it _looks_ like:
When a given module is imported from `node_modules`, it's `"ret"` starts with the drive letter "C:", but when it's imported _outside_ of that, it does not. I wonder if that's related.

avivkeller:
@nodejs/web-infra requesting fast track, as this is used in a reproduction for Rolldown.

PR #363, opened 2025-07-20
fix(win): use windows-safe paths correctly

shulaoda:
## Description

Let me briefly explain where the issue lies.

### Even when we use Windows-style paths, it still breaks.

The root cause is the following code: https://github.com/shulaoda/doc-kit/blob/10869bf6b60c1a41ebc188516c7d6a631d4dfef5/src/generators/web/utils/generate.mjs#L51-L72

```js
[
  'import NavBar from "D:\\shulaoda\\doc-kit\\src\\generators\\web\\ui\\components\\NavBar";',
  'import SideBar from "D:\\shulaoda\\doc-kit\\src\\generators\\web\\ui\\components\\SideBar";',
].join('')
```
After joining the strings, the `\\` sequences are interpreted as single backslashes (`\`), resulting in:

```js
'import NavBar from "D:\shulaoda\doc-kit\src\generators\web\ui\components\NavBar";import SideBar from "D:\shulaoda\doc-kit\src\generators\web\ui\components\SideBar";'
```
This causes the bundler to fail because the backslashes are treated either as escape characters or as invalid syntax.

### Why we need `toPosixPath`
There’s a problem with how we currently handled Windows paths. When `ROOT` is something like `D:/shulaoda/doc-kit/src/generators/web`, using `path.posix.resolve(ROOT, './ui/components/NavBar')` results in:

```shell
/shulaoda/doc-kit/D:/shulaoda/doc-kit/src/generators/web/ui/components/NavBar
```
This happens because `path.posix` doesn’t understand the `D:/` drive prefix — it treats it as a normal directory name and incorrectly prepends the current working directory (`/shulaoda/doc-kit`) to it.

## Validation

I verified it locally on my Windows.

## Related Issues

closes https://github.com/rolldown/rolldown/issues/5364

### Check List

### DIFF
--- a/src/generators/web/constants.mjs
+++ b/src/generators/web/constants.mjs
@@ -1,16 +1,17 @@
-import { parse, relative, sep, dirname } from 'node:path';
-import { resolve } from 'node:path/posix';
+import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-// Convert the current module's URL to a filesystem path,
-// then calculate the relative path from the system root directory
-// to this file. This relative path uses platform-specific separators,
-// so replace them with forward slashes ("/") for consistency and web compatibility.
-// Finally, prepend a leading slash to form an absolute root-relative path string.
-//
-// This produces a POSIX-style absolute path, even on Windows systems.
-const dir = dirname(fileURLToPath(import.meta.url));
-export const ROOT = '/' + relative(parse(dir).root, dir).replaceAll(sep, '/');
+export const ROOT = dirname(fileURLToPath(import.meta.url));
+
+/**
+ * Converts a filesystem path to POSIX-style format on non-POSIX platforms.
+ *
+ * @param {string} path - The original file path.
+ * @returns {string} The normalized path using POSIX-style separators if applicable.
+ */
+export const toPosixPath = path => {
+  return process.platform === 'win32' ? path.replaceAll('\\', '/') : path;
+};
 
 /**
  * @typedef {Object} JSXImportConfig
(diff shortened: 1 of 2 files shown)


src/generators/web/constants.mjs
```diff
@@ -1,16 +1,17 @@
...
 
-// Convert the current module's URL to a filesystem path,
-// then calculate the relative path from the system root directory
-// to this file. This relative path uses platform-specific separators,
-// so replace them with forward slashes ("/") for consistency and web compatibility.
-// Finally, prepend a leading slash to form an absolute root-relative path string.
-//
-// This produces a POSIX-style absolute path, even on Windows systems.
-const dir = dirname(fileURLToPath(import.meta.url));
-export const ROOT = '/' + relative(parse(dir).root, dir).replaceAll(sep, '/');
```
avivkeller, line 13: For reference, `relative(parse(dir).root, dir).replaceAll(sep, '/')` removes any drive letters

avivkeller: > If that's the case, Rolldown won't be able to recognize paths like `/shulaoda/dockit/src/generators/web/ui/components/NavBar` on Windows.

I believe it can. I've tested it. It successfully imports the paths, AFAIK. If you'd like, I can provide you with the trace logs?

> Could you tell me what the value of `ROOT` would be if it's set to `D:\\shulaoda\\doc-kit\\src\\generators\\web`, and what the result of `resolve(ROOT, './ui/components/NavBar')` would be?

```js
import { parse, relative, sep, dirname } from 'node:path';
import { resolve } from 'node:path/posix';

const dir = dirname("D:\\shulaoda\\doc-kit\\src\\generators\\web");
export const ROOT = '/' + relative(parse(dir).root, dir).replaceAll(sep, '/');

console.log('ROOT', ROOT);
console.log('RESOLVE', resolve(ROOT, './ui/components/NavBar'));
```
```
ROOT /shulaoda/doc-kit/src/generators
RESOLVE /shulaoda/doc-kit/src/generators/ui/components/NavBar
```

shulaoda: !!! Okay, I got it. That makes sense then — I’m now able to reproduce it as well.

PR #365, opened 2025-07-20
feat(web): improve deprecation page

avivkeller:
image_720

Improves the deprecation page by:
1. Not slicing the `DEPXXX:` off the title
2. Not bolding the code blocks, they should inherit their weight
3. Putting the deprecation type inside an AlertBox


AugustinMauroy (approved):
LGMT

PR #366, opened 2025-07-21
chore(queries): organize and test

### LINKED ISSUE #356: Headers with multiple `.`'s are not treated as valid methods.
Per our RegExp, ``/^`?(?:\w*(?:(\[[^\]]+\])|(?:\.(\w+)))|(\w+))\([^)]*\)`?$/i``, methods like `foo.bar.method()` are not valid.

avivkeller:
The queries and RegExp's we use should be extensively tested to prevent anything slipping through. This PR moves all of our RegExps and queries into their own file, and adds unit tests. This way, we can re-use the same testing structure across our RegExps, to prevent de-syncs.

### DIFF
--- a/src/generators/api-links/utils/extractExports.mjs
+++ b/src/generators/api-links/utils/extractExports.mjs
@@ -2,8 +2,6 @@
 
 import { visit } from 'estree-util-visit';
 
-import { CONSTRUCTOR_EXPRESSION } from '../constants.mjs';
-
 /**
  * @see https://github.com/estree/estree/blob/master/es5.md#assignmentexpression
  *
@@ -99,7 +97,7 @@ function handleExpression(node, basename, nameToLineNumberMap) {
             case 'Identifier': {
               exports.identifiers.push(value.name);
 
-              if (CONSTRUCTOR_EXPRESSION.test(value.name[0])) {
+              if (/^[A-Z]/.test(value.name)) {
                 exports.ctors.push(value.name);
               }
 
--- a/src/utils/queries/index.mjs
+++ b/src/utils/queries/index.mjs
@@ -15,9 +15,9 @@ import { getRemark } from '../remark.mjs';
 import { transformNodesToString } from '../unist.mjs';
 import {
   MARKDOWN_URL,
-  NORMALIZE_TYPES,
   STABILITY_INDEX,
   STABILITY_INDEX_PREFIX,
+  TYPE_EXPRESSION,
   UNIX_MANUAL_PAGE,
 } from './regex.mjs';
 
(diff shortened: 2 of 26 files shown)


AugustinMauroy (approved):
LGTM ! but could you provide a before after for the issue

avivkeller:
## After

Screenshot 2025-07-21 at 16 16 40

## Before

Screenshot 2025-07-21 at 16 16 47

AugustinMauroy:
Thanks 👍

araujogui (approved):

avivkeller:
Bump @nodejs/web-infra, this is a bug fix

ovflowd (changes requested):
> Bump @nodejs/web-infra, this is a bug fix

How is such a huge PR a bug fix? This seems a refactor per PR-description. What are you fixing?

src/utils/queries/index.mjs
```diff
@@ -13,6 +13,13 @@ import {
 } from '../parser/index.mjs';
 import { getRemark } from '../remark.mjs';
 import { transformNodesToString } from '../unist.mjs';
+import {
```
ovflowd, line 16: I'd rather a prefixed import

ovflowd:
> > How is such a huge PR a bug fix?
> 
> This PR refactors our RegExp queries, and, in the process, fixes #356, I should not have classified this as a "bug fix", it's a change that also fixes a bug.
> 
> > What I fail to understand is why you arbitrarily decided to redesign the structure of the whole queries
> 
> I figured, while I'm adding a RegExp file, why not move the query RegExps there as well, and at that point, _also_ move the Unist functions. But I'm happy to revert, since that was quite the tangent.

I do insist. Not against the refactor, but make a dedicated PR just for the refactor, without the bug fix.

This allows us to discuss the bug fix on this PR and the refactor on another :)

avivkeller:
Blocked by #379.

TODO Add test cases

avivkeller:
The RegExps have 100% test coverage, with around 600 lines of tests (but I'm open to adding more). Ideally, these RegExps will almost never change, lest we break something

src/utils/queries/index.mjs
ovflowd: Cleaner in this situation, these are a bunch on non prefixed constants, doing a wildcard import here better explains where they belong to.

src/generators/api-links/utils/extractExports.mjs
```diff
@@ -99,7 +97,7 @@ function handleExpression(node, basename, nameToLineNumberMap) {
             case 'Identifier': {
               exports.identifiers.push(value.name);
 
-              if (CONSTRUCTOR_EXPRESSION.test(value.name[0])) {
+              if (/^[A-Z]/.test(value.name)) {
```
ovflowd, line 100: Let's keep regexes in constants for easily finding regexes in one place

ovflowd: Possibly yes?

src/utils/queries/regex.mjs
```diff
@@ -0,0 +1,296 @@
+// ============================================================================
```
ovflowd, line 1: Instead of doing this, why not different files? These section blocks aren't a common thing in JavaScript.

avivkeller:
Okay, perhaps refactoring the RegExps wasn't really needed. I'm closing this.

avivkeller:
> I think my main issue with this PR is... What's the motivation behind it? Why is this refactoring needed? What's the problems with the current approach?

Oh! I didn't even see this comment, I suppose we reached the same conclusion :)

PR #381, opened 2025-07-27
feat(metabar): improve headings and add stability

avivkeller:
This PR adds Stability notices to the sidebar ("E" for Experimental, "L" for Legacy, "D" for Deprecated, and nothing for Stable). Additionally, it trims CLI flags and Environment variables down to just their names, similarly to how we do for methods.

image
image


AugustinMauroy (approved):
LGTM

PR #387, opened 2025-08-04
fix(ci): use PAT

avivkeller:
I made the silly little mistake of not setting the PAT 🤦

### DIFF
--- a/.github/workflows/update-internal-dependencies.yml
+++ b/.github/workflows/update-internal-dependencies.yml
@@ -50,4 +50,4 @@ jobs:
           branch: 'upgrade-node-core-deps'
           path: '.'
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


avivkeller:
@legendecas "Error: Resource not accessible by personal access token"

avivkeller:
Perhaps the token also needs Pull Request permissions?
https://github.com/nodejs/doc-kit/tree/upgrade-node-core-deps

avivkeller:
Oh wait, nevermind, I believe the error is because the PAT is including this PR's commits in it's changeset, which requires `actions` permissions.

cc @nodejs/web-infra fast track

MattIPv4 (approved):
Pretty sure it'll need the PR permission, but this needs to land either way

avivkeller:
> Pretty sure it'll need the PR permission, but this needs to land either way

Indeed, 🤦

PR #395, opened 2025-08-07
hotfix(rolldown): fix breaking change

avivkeller:
@nodejs/web-infra Requesting fast-track.

The recent rolldown upgrade performed by Dependabot contained a breaking change. This PR updates our codebase to work with the new rolldown version.


MattIPv4:
Why did no CI catch this? Can we make sure we have whatever we need in CI so we can trust dependency upgrades?

PR #414, opened 2025-08-19
fix(isTypedList): validate property syntax

avivkeller:
Previously, any `inlineCode` block inside a list that matched the syntax of a typed-list would be considered valid. After this, the `inlineCode`'s content _must_ also be a valid JavaScript property (`/^[a-z0-9$_]+$/i`[^1]). Additionally, previously, the node immediately following the property in a typed list was assumed to be a text node, however, in non-Typed lists, this could realisticly be any node, so a nullish check was added.

[^1]: Note: In this RegExp, properties can begin with numbers. This is intentional, since, in the event of an array, a number property may be valid in a typed list.

### DIFF
--- a/src/utils/queries/constants.mjs
+++ b/src/utils/queries/constants.mjs
@@ -4,4 +4,4 @@
 export const DOC_API_STABILITY_SECTION_REF_URL =
   'documentation.html#stability-index';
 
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):
Need to add test 😁

PR #436, opened 2025-09-03
feat(logger): use threaded logger

avivkeller:
This PR makes the logger rely on `winston`, a library used for multiple logger transports. Additionally, by bubbling all errors to the main thread, we can avoid the `getInstance()` method, and just export the logger directly.


ovflowd (commented):
I'm reluctant on approving this as it pretty much removes all the work @araujogui did. If we wanted to use a 3rd party, why all that effort?

PR #440, opened 2025-09-20
feat(css): responsive tables

avivkeller:
This PR bumps ui-components to include the responsive tables.

Additionally, it bumps all dependencies and adds `"eslint-import-resolver-node"`, since my local linting wasn't working without it, however, I can revert that if linting is unaffected in CI.

The downside to this is that sub-tables have very little space, shown below:

image


AugustinMauroy (approved):
LGMT !

PR #450, opened 2025-10-01
fix: add `@actions/core` as dependency

flakey5:
`@actions/core` is used in the GitHub logger transport despite not actually being installed directly, meaning it's only available when `@reporters/github` is downloaded w/ dev dependencies. In Node core we don't install dev dependencies causing this to throw.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,6 +6,7 @@
     "": {
       "name": "@nodejs/doc-kit",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


flakey5:
requesting fast track since blocks node core pr

MattIPv4 (approved):

MattIPv4:
smth smth pnpm and this wouldnt be a problem

PR #479, opened 2025-11-03
fix: add legacy-html-assets generator

flakey5:
Move asset folder copying to its own generator that we can call separately from the legacy-html generator

### Check List


AugustinMauroy (approved):

avivkeller:
I don't love this approach, it feels like a waste of a generator. This should, IMO, be one generator, or no generators and handled by the Makefile (although I'd prefer the former)

PR #512, opened 2025-12-11
feat(orama): use new ui components

avivkeller:
This PR uses the new `@orama/ui` library

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -11,7 +11,7 @@
         "@heroicons/react": "^2.2.0",
         "@minify-html/node": "^0.16.4",
         "@node-core/rehype-shiki": "1.3.0",
-        "@node-core/ui-components": "^1.4.1",
+        "@node-core/ui-components": "1.4.1",
         "@orama/orama": "^3.1.16",
         "@orama/ui": "^1.5.3",
         "@rollup/plugin-virtual": "^3.0.2",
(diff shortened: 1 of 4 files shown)


src/generators/web/ui/index.css
```diff
@@ -70,3 +70,17 @@ main {
...
+  }
+  > div {
+    min-width: 0;
+    > p {
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+}
```
avivkeller: Note: This can technically be done in `@node-core/ui-components`, would we prefer that?

It'll have no effect on the website, except for that it'll create unused CSS code.

src/generators/web/ui/hooks/useOrama.mjs
```diff
@@ -13,6 +13,12 @@ export default () => {
       schema: {},
     });
 
+    // TODO(@avivkeller): Ask Orama to support this functionality natively
+    /**
+     * @param {any} options
+     */
+    db.search = options => search(db, options);
+
```
avivkeller: cc @micheleriva: `@orama/ui` calls `client.search` (https://github.com/oramasearch/orama-ui/blob/main/packages/ui/src/hooks/useSearch.tsx#L94), however, the non-Cloud client don't implement this function. Instead, they use `search(client, ...)` syntax.

Is there a way for Orama to add `client.search()` functionality to the non-Cloud clients, **or** support `search(client, ...)` in the UI library?

PR #517, opened 2025-12-11
feat(ci): compare builds

avivkeller:
This PR adds a system for adding a series of comparators to our codebase.

At the moment, only a `web` comparison is implemented. To add more comparators, add `[generator-name].mjs` to the `scripts/compare-builds` directory. These comparators will run on incoming PRs.

Example output: https://github.com/nodejs/doc-kit/pull/517#issuecomment-3644423461

### DIFF
--- a/.github/workflows/compare-builds.yml
+++ b/.github/workflows/compare-builds.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Download Output (BASE)
         uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
         with:
-          name: web
+          name: ${{ matrix.comparator }}
           path: out/base
           run-id: ${{ steps.base-run.outputs.run_id }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


src/generators/web/utils/processing.mjs
```diff
@@ -118,7 +117,7 @@ export async function processJSXEntries(
       .replace('{{dehydrated}}', serverBundle.get(fileName) ?? '')
       .replace('{{importMap}}', clientBundle.importMap ?? '')
       .replace('{{entrypoint}}', `./${fileName}?${randomUUID()}`)
-      .replace('{{speculationRules}}', speculationRulesString);
```
avivkeller: **Note**: This change lowers the build size a bit, which helps with the demo comment. I can revert if needed.

.github/workflows/compare-builds.yml
```diff
@@ -0,0 +1,125 @@
...
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: ${{ matrix.comparator }}
+          path: result.txt
+
+  aggregate:
+    name: Aggregate Comparison Results
+    runs-on: ubuntu-latest
+    needs: compare
+    steps:
```
flakey5: Should we have the harden runner step here?

PR #520, opened 2025-12-15
feat: create sitemap generator

### LINKED ISSUE #255: Generate sitemap.xml in the generator

We should have the ability to generate the API sitemap. Currently, there is no generated sitemap, which could cause new APIs to be indexed by bots like Google, showing outdated information (see https://github.com/nodejs/nodejs.org/issues/6577).

araujogui:
## Description

Create sitemap.xml generator

## Related Issues

Fixes #255

### Check List

### DIFF
--- a/src/generators/sitemap/index.mjs
+++ b/src/generators/sitemap/index.mjs
@@ -1,8 +1,6 @@
 import { readFile, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
 
-import dedent from 'dedent';
-
 import { BASE_URL } from '../../constants.mjs';
 import { createPageSitemapEntry } from './utils/createPageSitemapEntry.mjs';
 
@@ -30,17 +28,29 @@ export default {
    * @returns {Promise<string>}
    */
   async generate(entries, { output }) {
+    const template = await readFile(
+      join(import.meta.dirname, 'template.xml'),
+      'utf-8'
+    );
+
+    const entryTemplate = await readFile(
+      join(import.meta.dirname, 'entry-template.xml'),
+      'utf-8'
+    );
+
     const lastmod = new Date().toISOString().split('T')[0];
 
     const apiPages = entries
       .filter(entry => entry.heading.depth === 1)
       .map(entry => createPageSitemapEntry(entry, lastmod));
 
+    const { href: loc } = new URL('/docs/latest/api/', BASE_URL);
+
     /**
      * @typedef {import('./types').SitemapEntry}
      */
     const mainPage = {
-      loc: new URL('/docs/latest/api/', BASE_URL).href,
+      loc,
       lastmod,
       changefreq: 'daily',
       priority: '1.0',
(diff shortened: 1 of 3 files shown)


avivkeller (commented):
Thanks! I know this is a draft, but a few notes that I thought I might share. 

These aren't blockers or concerns, just little notes.

src/generators/sitemap/index.mjs
```diff
@@ -0,0 +1,75 @@
...
+  async generate(entries, { output }) {
+    const apiPages = entries
+      .filter(entry => entry.heading.depth === 1)
+      .map(entry => {
+        const path = entry.api_doc_source.replace(/^doc\//, '/docs/latest/');
+        const url = new URL(path, BASE_URL).href;
+
+        return {
+          loc: url,
+          lastmod: new Date().toISOString().split('T')[0],
```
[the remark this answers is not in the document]
araujogui: Actually, I was thinking of somehow getting the last modification date from the git repository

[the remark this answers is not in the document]
avivkeller: Hmm, that could also work, neat!

araujogui:
> @araujogui is this ready for review?

Not yet, I still need to implement: https://github.com/nodejs/doc-kit/pull/520#discussion_r2621310687

src/generators/sitemap/index.mjs
```diff
@@ -0,0 +1,81 @@
...
+   * @param {Input} entries
+   * @param {Partial<GeneratorOptions>} options
+   * @returns {Promise<string>}
+   */
+  async generate(entries, { output }) {
+    const lastmod = new Date().toISOString().split('T')[0];
+
+    const apiPages = entries
+      .filter(entry => entry.heading.depth === 1)
+      .map(entry => {
```
ovflowd: nit: extract this anonymous function to a dedicated function.

src/generators/sitemap/index.mjs
```diff
@@ -0,0 +1,81 @@
...
+        const path = entry.api_doc_source.replace(/^doc\//, '/docs/latest/');
+        const url = new URL(path, BASE_URL).href;
```
ovflowd:
```suggestion
        const { href: url } = new URL(path, BASE_URL);
```

ovflowd: nit: prefer object deconstruction.

src/generators/sitemap/index.mjs
```diff
@@ -0,0 +1,81 @@
...
+
+    const template = await readFile(
+      join(import.meta.dirname, 'template.xml'),
+      'utf-8'
+    );
+
+    const urlset = apiPages
+      .map(
+        page => dedent`
+        <url>
```
ovflowd: nit: use template files for this and then do simple key->value substitution. Or use proper rss/feed libraries OR xml libraries.

src/generators/sitemap/index.mjs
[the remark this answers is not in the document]
ovflowd: Im fine with either :)

src/generators/sitemap/index.mjs
ovflowd: I'm fine with whatever that uses the least amount of dependencies. We can also just use yet another template file, we can also simply use another dependency, like the xast one, or rss/feeds or whatever.

src/utils/url.mjs
```diff
@@ -0,0 +1,19 @@
+import { BASE_URL } from '../constants.mjs';
+
+/**
+ * Builds the url of a api doc entry.
+ *
+ * @param {ApiDocMetadataEntry} entry
+ * @param {boolean} [useHtml]
+ * @returns {URL}
+ */
+export const buildApiDocURL = (entry, useHtml = false) => {
```
avivkeller:
```suggestion
export const buildApiDocURL = (entry) => {
```

Let's always use HTML until we have a need not to

src/generators/sitemap/types.d.ts
```diff
@@ -0,0 +1,13 @@
+export interface SitemapUrl {
```
avivkeller:
```suggestion
export interface SitemapEntry {
```

src/generators/sitemap/index.mjs
```diff
@@ -0,0 +1,77 @@
...
```
ovflowd: nit: this should be also another template string coming from a file.

src/utils/url.mjs
```diff
@@ -0,0 +1,19 @@
...
+  const path = entry.api_doc_source.replace(/^doc\//, '/docs/latest/');
+
+  if (useHtml) {
+    const htmlPath = path.replace(/\.md$/, '.html');
```
ovflowd: nit: no need to assign const htmlPath just return new URL();

Also instead of using new URL() on all these places, could you use URL.parse please?

PR #524, opened 2025-12-19
fix(legacy-json): reduce diff

avivkeller:
Ref: https://gist.github.com/avivkeller/6bcb5fc17d72a3bed77744d4094949bd

The changes are backwards compared to the legacy implementation.


avivkeller:
**Note**: The comparator could not run for this PR, as the diff (20k+ lines) was simply too large for a comment:
https://github.com/nodejs/doc-kit/actions/runs/20384428064/job/58582289227#step:6:4

Issue #535, opened 2025-12-25
General UI/UX improvements


canerakdas:
- On mobile layouts, headings are getting cut off. The design could be updated to better match Figma.

<img width="430" height="381" alt="Image" src="https://github.com/user-attachments/assets/56eac4a8-c4c8-4e96-8a1e-d5dda0c1c894" />

Figma;
<img width="430" height="381" alt="Image" src="https://github.com/user-attachments/assets/3bb749b8-b9b5-421f-bc98-67a433976186" />

https://www.figma.com/design/a10cjjw3MzvRQMPT9FP3xz/Node.js?node-id=696-19108&t=1LYJN1JiPyrfM4Ie-0

- In the documentation, the second select in the root directory shows up empty. Adding a label or placeholder would make it clearer even when nothing is selected.

<img width="430" height="381" alt="Image" src="https://github.com/user-attachments/assets/e4d1e57a-c52a-4610-95b9-5a5ada56dee4" />

- I think the tab titles are not very clear. They actually indicate the type of code shown and are essentially examples. Renaming them to something like Example [CJS], Example [MJS], and displaying the type separately could be more intuitive.

<img width="771" height="334" alt="Image" src="https://github.com/user-attachments/assets/153e6fd0-ce37-4dfd-b435-21f5da6d68ce" />

For example;
<img width="774" height="159" alt="Image" src="https://github.com/user-attachments/assets/4ff7e1a8-ffae-48d9-bdd0-fa9a22d7d1f8" />

- The colors of tables and code boxes don’t match. Using similar border colors would improve visual consistency.

<img width="774" height="361" alt="Image" src="https://github.com/user-attachments/assets/3d51c4c3-533f-4a75-a596-2225af52e373" />

- Meta description tags are missing on the pages; we could add them (possibly reusing the ones from the website).

<img width="592" height="94" alt="Image" src="https://github.com/user-attachments/assets/e5fab0bb-1030-4b1b-bcb4-20c0a3c2817b" />

- Adding a simple footer could be useful, including important links for us (even if they’re not things like Latest LTS or Latest Release).

- For missing values in tables, we currently use “-”. If a column doesn’t have values at all, removing that column entirely might result in a mobile/desktop cleaner appearance.

<img width="814" height="267" alt="Image" src="https://github.com/user-attachments/assets/ba644afd-8318-4001-bb70-244fb6ec2c95" />

- We could reuse the OG image from the website here as well, to ensure a more consistent look when shared on social media.

[...]

PR #538, opened 2025-12-26
feat: og image and meta description

canerakdas:
## Description

With this PR, we aim to use the OG images / meta descriptions from the website with the new web generator as well

## Validation

In the preview should contain og image content and display it correctly

## Related Issues

#535 

### Check List

### DIFF
--- a/src/generators/web/utils/processing.mjs
+++ b/src/generators/web/utils/processing.mjs
@@ -112,7 +112,6 @@ export async function processJSXEntries(
   const results = entries.map(({ data: { api, heading } }) => {
     const fileName = `${api}.js`;
     const title = `${heading.data.name} | ${titleSuffix}`;
-    const description = `Documentation for ${heading.data.name} in Node.js v${version.version}`;
 
     // Replace template placeholders with actual content
     const renderedHtml = template
@@ -121,8 +120,7 @@ export async function processJSXEntries(
       .replace('{{importMap}}', clientBundle.importMap ?? '')
       .replace('{{entrypoint}}', `./${fileName}?${randomUUID()}`)
       .replace('{{speculationRules}}', SPECULATION_RULES)
-      .replace('{{ogTitle}}', title)
-      .replace('{{ogDescription}}', description);
+      .replace('{{ogTitle}}', title);
 
     // Minify HTML (input must be a Buffer)
     const finalHTMLBuffer = HTMLMinifier.minify(Buffer.from(renderedHtml), {});


src/generators/web/template.html
```diff
@@ -6,7 +6,11 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <link rel="icon" href="https://nodejs.org/static/images/favicons/favicon.png"/>
   <title>{{title}}</title>
+  <meta name="description" content="Node.js® is a free, open-source, cross-platform JavaScript runtime environment that lets developers create servers, web apps, command line tools and scripts.">
```
canerakdas: I'm open to suggestions for a better description

bjohansebas:
> Why? We already have it on our website, why add complexity here?

Maybe because if the URL changes (which I hope never happens), the OG image would no longer work. By generating it here, that problem wouldn’t exist.

canerakdas:
> > Why? We already have it on our website, why add complexity here?
> 
> Maybe because if the URL changes (which I hope never happens), the OG image would no longer work. By generating it here, that problem wouldn’t exist.

I think we also cannot use these images in a scenario where the website is statically exported, but I agree that there is a cost to creating them

src/generators/web/template.html
AugustinMauroy: So it's good 👍

canerakdas:
Looks like we have two possible approaches here;

By default, we can use the dynamic version we use on Nodejs.org, and fall back to the static version [used on the homepage](https://nodejs.org/en/next-data/og/announcement/Node.js%20%E2%80%94%20Run%20JavaScript%20Everywhere) if that dynamic image isn’t available (website created statically). This approach depends on the website’s Open Graph images, so it should be properly documented in both repo.

Alternatively, we could build an OG image generator within `doc-kit`. I put together a small example over the weekend using `@resvg/resvg-js` to generate images dynamically, but I don’t think this is a very efficient approach and it adds an extra layer of complexity.

@nodejs/nodejs-website @nodejs/web-infra  WDYT?

canerakdas:
> Do we really need og generation for api docs? I mean are docs from other platforms using generic og for their docs or one per page?

Sorry for the late reply 🙇 

From what I’ve seen, most platforms use a generic image featuring their mascot. Python is the exception, where a unique Open Graph image is generated per page

image

IMO it looks nice, but I’m not convinced it’s something we need to ship from day one. For now, I’m leaning toward going with a generic image (since it’s better than nothing and simple), unless there’s a strong opposition 🙌

AugustinMauroy:
IMO:
- We can start by just one image
- If we can have something like python or next.js that will be awesome !

canerakdas:
Here is the example output;

image

ovflowd (approved):
SGTM!

PR #554, opened 2025-12-27
Fix(web): hide empty navbar space on mobile

### LINKED ISSUE #533: Mobile Menu with weird paddings
Because doc-kit `web` generator has no menu entries on navigation header, we have a "chunk" of empty space, we should probably hide that if there are no entries.

Image

arvindm84:
## Description

This PR fixes issue #533 where the mobile navigation header shows a chunk of empty space when there are no navigation menu entries. The web UI now hides the empty space in the navbar. It also ensures the web CSS bundling step consistently preserves the generated CSS during bundling, and contains tests to cover the CSS bundling behavior.

## Validation

Before:
Before

After:
After

## Related Issues

Fixes #533 

### Check List

- [x] I have checked code formatting with `node --run format` & `node --run lint`.

### DIFF
--- a/src/generators/api-links/__tests__/fixtures.test.mjs
+++ b/src/generators/api-links/__tests__/fixtures.test.mjs
@@ -39,7 +39,7 @@ describe('api links', () => {
         const astJsResults = [];
 
         for await (const chunk of astJs.generate(undefined, {
-          input: [sourceFile],
+          input: [sourceFile.replaceAll('\\', '/')],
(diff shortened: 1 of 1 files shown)


avivkeller (changes requested):
Thanks for your PR!

You should be able to achieve this with CSS-only changes.

Note that the `ui` directory should already have a css file (or it should be somewhere nearby)

PR #564, opened 2026-01-14
meta: bump ui-components

btea:
## Description
continue https://github.com/nodejs/doc-kit/pull/562#pullrequestreview-3634378082

### Check List


avivkeller (changes requested):
Let’s wait for Dependabot

btea:
How long is the Dependabot update cycle?

Moreover, it seems that simply updating the component library is not enough, it appears that related logic also needs to be added.

ovflowd:
Given that, I appreciate your contributions @btea, but we won't accept a manual version bump PR at this time, thank you!

Issue #568, opened 2026-01-27
Automate npm publication as @node-core/doc-kit with version bump workflow


ovflowd:
### Enter your suggestions in details:

**Goal:** Publish the `doc-kit` package on npm as `@node-core/doc-kit`.

**Automation:** Use a similar workflow for automatically publishing the package on a new version bump as implemented for `@node-core/ui-components` in the `@nodejs/nodejs.org` repository.

**Details:**
- Automate the npm publish process so that every time a new version of `@node-core/doc-kit` is released, it is automatically published to npm.
- Use GitHub Actions to implement this automation, leveraging the workflow already established for `@node-core/ui-components` in `@nodejs/nodejs.org` as a reference.
- The [`publish.mjs` script in ui-components](https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/scripts/publish.mjs) generates a cleaned `package.json` in `dist/`, copies documentation, and runs `pnpm publish --no-git-checks` in the `dist` directory.
- The workflow is set up so publishing is triggered on merges to `main` (after CI success) or via `workflow_dispatch`. More info and diagrams: [docs/package-publishing.md](https://github.com/nodejs/nodejs.org/blob/main/docs/package-publishing.md).
- The monorepo structure supports independent package publishing managed by GitHub Actions.
- Document any secrets (npm tokens), configuration steps, or permission requirements for publishing—including any changes needed for branch protection, secrets, or CI.

**Plan:**
1. Review [`ui-components` publish script](https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/scripts/publish.mjs) and [docs](https://github.com/nodejs/nodejs.org/blob/main/docs/package-publishing.md).
2. Adapt a similar workflow and script for use in `doc-kit` (ensure it strips devDependencies, formats paths, and publishes from `dist/`).
3. Create a GitHub Action to trigger the publish job after a version bump, following merge to main or manual trigger—ensure it depends on the successful completion of tests.
4. Test by bumping `@node-core/doc-kit` version and verifying an automated npm publish.
5. Document the setup and maintenance steps in the repository README.

***References:***
- [`publish.mjs` script for ui-components](https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/scripts/publish.mjs)
- [Publishing process documentation](https://github.com/nodejs/nodejs.org/blob/main/docs/package-publishing.md)
[...]

PR #593, opened 2026-02-07
feat(cli): add a configuration file option

avivkeller:
Adds support for a `--config-file` option, allowing for custom configuration files to be passed. These configuration files must be loadable via a `import()` call, so usually JSON or JavaScript files with default exports.

The `global` object contains settings that apply to all generators unless overridden:

| Property     | Type               | Description                                | Default                                            |
| ------------ | ------------------ | ------------------------------------------ | -------------------------------------------------- |
| `version`    | `string \| SemVer` | Documentation version                      | `process.version`                                  |
| `minify`     | `boolean`          | Whether to minify output                   | `false`                                            |
| `repository` | `string`           | GitHub repository in `owner/repo` format   | `'nodejs/node'`                                    |
| `ref`        | `string`           | Git reference (branch, tag, or commit SHA) | `'HEAD'`                                           |
| `baseURL`    | `string \| URL`    | Base URL for documentation                 | `'https://nodejs.org/docs'`                        |
| `input`      | `string[]`         | Input directory path                       | -                                                  |
| `output`     | `string`           | Output directory path                      | -                                                  |
| `ignore`     | `string[]`         | Patterns to ignore                         | `[]`                                               |
| `changelog`  | `string \| URL`    | Changelog URL                              | Auto-generated URL based on `ref` and `repository` |
| `index`      | `string \| URL`    | Index URL                                  | -                                                  |

Each generator (e.g., `json`, `html`, `markdown`) can have its own configuration that overrides global settings:

```javascript
export default {
  global: {
    version: '20.0.0',
    minify: true,
  },

  'legacy-json': {
    minify: false, // Override: JSON output won't be minified
  },
};
```

Configurations are merged in the following order (earlier sources take precedence):

1. **Config file** (`--config-file`)
2. **CLI options** (command-line arguments)
3. **Default values** (built-in defaults)

CLI options map to configuration properties:

| CLI Option             | Config Property    | Example                   |
| ---------------------- | ------------------ | ------------------------- |
| `--input <path>`       | `global.input`     | `--input src/`            |
| `--output <path>`      | `global.output`    | `--output dist/`          |
| `--ignore <pattern>`   | `global.ignore[]`  | `--ignore test/`          |
| `--minify`             | `global.minify`    | `--minify`                |
| `--git-ref <ref>`      | `global.ref`       | `--git-ref v20.0.0`       |
| `--version <version>`  | `global.version`   | `--version 20.0.0`        |
| `--changelog <url>`    | `global.changelog` | `--changelog https://...` |
| `--index <url>`        | `global.index`     | `--index file://...`      |
| `--type-map <map>`     | `metadata.typeMap` | `--type-map file://...`   |
| `--target <generator>` | `target`           | `--target json`           |
| `--threads <n>`        | `threads`          | `--threads 4`             |
| `--chunk-size <n>`     | `chunkSize`        | `--chunk-size 10`         |


avivkeller:
@nodejs/web-infra Once this lands, I think we can begin the monorepo work @ovflowd mentioned. Plus, if we combine that work with the publishing script in #568, we could even use typescript!

src/utils/configuration/__tests__/index.test.mjs
```diff
@@ -0,0 +1,254 @@
...
+        createMockConfig({ global: { input: 'custom-src/' } })
+      );
+
+      const config = await createRunConfiguration({
+        configFile: 'config.mjs',
+        output: 'custom-dist/',
+        threads: 2,
+      });
+
+      console.log(config);
```
ovflowd: @avivkeller you forgot... this. More self review is needed. I know this was merged, but can you please go through this whole diff and see if you missed anything and maybe do a cleanup of certain things?

PR #595, opened 2026-02-07
fix: filter out new keys from legacy JSON generator

aduh95:
### Check List


ovflowd (changes requested):
Let's wait for the TSC meeting.

src/utils/generators.mjs
```diff
@@ -109,3 +109,63 @@ export const sortChanges = (changes, key = 'version') => {
...
+  miscs,
+  modules,
+  globals,
+}) =>
+  JSON.stringify(
+    {
+      type,
+      source,
+      introduced_in,
+      ...(api === 'report'
```
ovflowd: @aduh95 can you add inline comment explaining why this different order with api === report?

PR #600, opened 2026-02-10
fix: refine FUNCTION_CALL regex for better matching

aduh95:
## Description

That should exclude https://nodejs.org/api/stream.html#readableread0 and https://nodejs.org/api/stream.html#readablepush to be parsed as methods.

### Check List


avivkeller (approved):

avivkeller (changes requested):
What about props that are optional?

PR #602, opened 2026-02-15
meta: switch to minify-html; fixed test glob on package.json; removed console.log

ovflowd:
This PR moves from @swc/html as not being supported on all platforms we need; It also fixes the glob pattern for test scripts, if you had out/ folder tests would fail lol; and removes some console.logs() from tests.


ovflowd:
cc @nodejs/web-infra if we could fast-track this to unblock nodejs/node that'd be great!

avivkeller:
We could do a fallback, ie try a binding and if not don't minify?

avivkeller (approved):

PR #610, opened 2026-02-16
fix(slugger): optimize trailing/leading regex

avivkeller:
Fixes https://openjs-foundation.slack.com/archives/C09EXEEHFKP/p1771022712509279

### DIFF
--- a/src/utils/parser/constants.mjs
+++ b/src/utils/parser/constants.mjs
@@ -5,7 +5,9 @@ export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
   { from: /&/, to: '-and-' }, // Replace &
   { from: /[/_,:;\\ ]/g, to: '-' }, // Replace /_,:;\. and whitespace
-  { from: /^-(?!$|-)|(?<!^|-)-$/g, to: '' }, // Remove any trailing or leading hyphen
+  { from: /^-+(?!-*$)/g, to: '' }, // Remove any leading hyphens
+  { from: /(?<!^-*)-+$/g, to: '' }, // Remove any trailing hyphens
+  { from: /^(?!-+$).*?(--+)/g, to: '-' }, // Replace multiple hyphens
 ];
 
 // This is the base URL of the MDN Web documentation


avivkeller:
**Note**: The anchors are _slightly_ different (i.e. `--my-flag`'s anchor is `--my-flag` instead of `-my-flag`)

src/utils/parser/constants.mjs
```diff
@@ -5,9 +5,7 @@ export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
   { from: /&/, to: '-and-' }, // Replace &
   { from: /[/_,:;\\ ]/g, to: '-' }, // Replace /_,:;\. and whitespace
-  { from: /--+/g, to: '-' }, // Replace multiple hyphens with single
-  { from: /^-/, to: '' }, // Remove any leading hyphen
-  { from: /-$/, to: '' }, // Remove any trailing hyphen
+  { from: /^-(?!$|-)|(?<!^|-)-$/g, to: '' }, // Remove any trailing or leading hyphen
```
MattIPv4:
```suggestion
  { from: /^-+(?!-*$)/g, to: '' }, // Remove any leading hyphens
  { from: /(?<!^-*)-+$/g, to: '' }, // Remove any trailing hyphens
```
I think this does the same, but is a bit clearer to me, matching one or more hyphens, as long as we don't see the start/end around.

src/utils/parser/constants.mjs
```diff
@@ -5,9 +5,7 @@ export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
   { from: /&/, to: '-and-' }, // Replace &
   { from: /[/_,:;\\ ]/g, to: '-' }, // Replace /_,:;\. and whitespace
-  { from: /--+/g, to: '-' }, // Replace multiple hyphens with single
```
MattIPv4: I think we probably want to keep this deduplication somehow?

MattIPv4: `(?<!^-*)--+(?!-*$)` might work for this?

MattIPv4 (approved):
LGTM, but I think this would benefit from some explicit test cases, given the complexity of this slugger logic (not just this change, but all the replacements happening).

PR #611, opened 2026-02-18
fix: use wasm

ovflowd:
This PR solves downstream failures due to lack of Rust, by using LightningCSS WASM version. It also fixes the dependency pinning using safer versions.


ovflowd:
> Regardless of there relevance to the project, they are irrelevant to this PR

No, they are relevant, or at least I deem them relevant. We can disagree on that 🤷

PR #613, opened 2026-02-18
meta: pinning dependencies based on guidelines

ovflowd:
This PR pins dependencies to ranges based on our [dependency pinning guidelines](https://github.com/nodejs/nodejs.org/blob/main/docs/dependency-pinning.md), it also updates rolldown, @node-core/ui-components and preact to latest sub-versions of their RCs.


MattIPv4:
I realise this is our policy currently, but I really do not see any value in this when we have a lockfile, and Dependabot will ignore this pinning anyway.

ovflowd:
cc @araujogui

PR #615, opened 2026-02-18
fix(rolldown): dynamic import

avivkeller:
This prevents non-`web` generators from throwing errors on machines that don't support the new generators. 

In the future, we will need a permanent solution (likely just not generating on these minority distros), but for now, this is a low-impact fix.

### DIFF
--- a/src/generators/web/utils/bundle.mjs
+++ b/src/generators/web/utils/bundle.mjs
@@ -1,11 +1,11 @@
 import { join } from 'node:path';
 
 import virtual from '@rollup/plugin-virtual';
-import { build } from 'rolldown';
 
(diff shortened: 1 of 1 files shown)


ovflowd (approved):

ovflowd (changes requested):
Wait, isn't the web generator already lazily imported? Why is it throwing error then?

AugustinMauroy (approved):

ovflowd:
> @ovflowd We load all the generators on the main thread to extract the configuration options. They are dynamically loaded on threads

Im fairly sure I changed the generators list so that the values only load when called. Why do we load all generators on main thread? We shouldn't. All we do is access the keys.

PR #616, opened 2026-02-19
meta: move generators

ovflowd:
This PR moves the generator strategy from lazy load + loading all at entry, to simply having the metadata be synchronous and always exposed and the actual generator implementation to be within a new file of generator.mjs for each generator. This reduces a bit asynchronous overload, separates schema from implementation and also ensures that generator dependencies and code gets initialized only when needed, which solves for now the downstream nodejs/node build.

### DIFF
--- a/docs/generators.md
+++ b/docs/generators.md
@@ -32,13 +32,14 @@ A generator is defined as a module exporting an object conforming to the `Genera
 
 ## Creating a Basic Generator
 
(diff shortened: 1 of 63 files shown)


ovflowd:
@avivkeller just do whatever with the PR, I don't have the time/mental availability to address your nitpicks now.

PR #637, opened 2026-03-03
feat(ui): use FunctionSignature component

avivkeller:
cc @canerakdas

---

This PR changes the `createPropertyTable` function into a function that generates a `FunctionSignature` component layout

image

### DIFF
--- a/src/generators/jsx-ast/utils/__tests__/signature.test.mjs
+++ b/src/generators/jsx-ast/utils/__tests__/signature.test.mjs
@@ -41,7 +41,7 @@ describe('generateSignature', () => {
         ''
       );
 
-      assert.strictEqual(sig, 'baz()');
+      assert.strictEqual(sig, 'baz(): void');
     });
 
     it('handles empty return type', () => {
@@ -54,7 +54,7 @@ describe('generateSignature', () => {
         ''
       );
 
-      assert.strictEqual(sig, 'test()');
+      assert.strictEqual(sig, 'test(): void');
     });
 
     it('includes prefix when provided', () => {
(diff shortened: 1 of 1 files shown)


canerakdas (approved):
Looks great, thank you Aviv! I'm testing the design across different pages, but from an implementation standpoint, everything looks good to me 🐢

src/generators/jsx-ast/utils/buildContent.mjs
```diff
@@ -21,7 +20,11 @@ import {
   TYPES_WITH_METHOD_SIGNATURES,
   TYPE_PREFIX_LENGTH,
 } from '../constants.mjs';
-import insertSignature, { getFullName } from './buildSignature.mjs';
+import {
+  insertSignatureCodeBlock,
```
ovflowd: name of fn is misleading, `insertSignatureSection`

src/generators/web/ui/index.css
```diff
@@ -16,33 +16,6 @@ main {
     word-break: break-word;
   }
 
-  /* Don't overflow the parent */
```
ovflowd: Why did you remove this? This was intentionally added by me...

ovflowd: I'm kinda challenging the habit of unrelated removals/changes that aren't really documented or explained anywhere. Let's please STOP touching the source code arbitrarily or at least at reasoning/comments on changes that seem unrelated.

src/generators/jsx-ast/utils/buildContent.mjs
avivkeller: That function is for the code block

ovflowd: Thanks for clarifying!

ovflowd:
image

Shouldn't in this case it be considered a constructor? maybe not cause of no `new` qualifier, but I think the signature should be there before any text.

image

Still seeing cases of only return

PR #641, opened 2026-03-03
fix(deps): downgrade preact

avivkeller:
I've made two discoveries in the past few minutes:
1. We have been using Preact's beta, which doesn't follow Semver, and makes Preact unupdatable

---

Requesting fast-track


MattIPv4 (changes requested):
There is a really significant diff on the lockfile, including what looks to be a very large number of transitive dependency bumps which don't seem related to downgrading preact?

What is the history of why this uses shrinkwrap instead of a lockfile, as I imagine that was a very intentional decision, and I'd like to understand that before we move forward with undoing that decision?

MattIPv4:
Per https://docs.npmjs.com/cli/v11/configuring-npm/npm-shrinkwrap-json, shrinkwarp is _exactly_ the same as lockfile other than that it gets published with the package, so I do not see why regeneration + bumping versions was needed, it should just be a case of renaming the file?

ovflowd (changes requested):
We should keep npm-shrinkwrap. We're not installing doc-kit via git. npm-shrinkwrao was a requirement done by @aduh95 -- I'd appreciate if we don't keep reverting previous decisions then doing fast-track on such quick decisions. Ask questions first, please (check git blame, for example, understand why certain decisions were made instead of quickly challenging them). Thanks!

ovflowd:
Even if it is not, there is no reason to go to package-lock. Sorry, for me this is a no. Not going to withdraw my block till better info is provided here, sorry Aviv 😭😅

AugustinMauroy (approved):
LGMT ! but for the fast-track wait Antoine to answer

PR #644, opened 2026-03-05
feat(`web`): introduce stability overview

### LINKED ISSUE #357: Generating the Stability Overview
Currently, in @flakey5's PR to node core, we generate each file individually (i.e. instead of passing `*.md` as an input, we run the command for _each_ file) as to increase speed via Makefile's caching.

However, because of this, we don't generate a Stability Overview on the "About" page.

AugustinMauroy:
## Description

introduce stability overview on new web generator

## Validation

It's should be cool 😎

## Related Issues

Close #357 

### Check List


src/generators/web/ui/components/StabilityOverview/index.jsx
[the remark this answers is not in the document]
AugustinMauroy: done !

avivkeller (commented):
I wonder if adding stability to the sidebar would be better?

src/generators/jsx-ast/generate.mjs
```diff
@@ -54,14 +59,30 @@ export async function* generate(input, worker) {
     ? config.index.map(({ section, api }) => [section, `${api}.html`])
     : headNodes.map(node => [node.heading.data.name, `${node.api}.html`]);
 
+  // Pre-compute stability overview data once — avoid serialising full AST nodes to workers
+  const stabilityOverviewEntries = headNodes
+    .filter(node => node.stability?.children?.length)
+    .map(({ api, heading, stability }) => {
+      const [{ data }] = stability.children;
+      return {
```
ovflowd: Could this map fn be moved to a dedicated function and be unit tested? Thanks!

ovflowd (changes requested):
This method would make sense for legacy docs. For the new docs, this method isn't really acceptable. The stability overview must be a React component now for the web generator, and all that we must do here is replace the STABILITY_OVERVIEW_SLOT_BEGIN or IDK with a slot for the component.

This would also make it easier to maintain the component.

src/generators/jsx-ast/generate.mjs
```diff
@@ -54,14 +59,30 @@ export async function* generate(input, worker) {
     ? config.index.map(({ section, api }) => [section, `${api}.html`])
     : headNodes.map(node => [node.heading.data.name, `${node.api}.html`]);
 
+  // Pre-compute stability overview data once — avoid serialising full AST nodes to workers
+  const stabilityOverviewEntries = headNodes
+    .filter(node => node.stability?.children?.length)
```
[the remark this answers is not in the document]
AugustinMauroy: oh 🤓 that's smart

ovflowd:
> I asked for an AST approach for generating a table, however, if that's not possible for a sidebar, we don't have to do it.
> 
> I meant, the badge that you have in the table might be better suited in the sidebar, WDYT?

I understand why an AST approach, but I'd prefer a React component here, we're using AST trees for building the structure of the page, which is fine I guess? But I'd rather have it being a React tree, not gonna lie. For example a lot of things are done in places that don't make much sense for me, for example https://github.com/nodejs/doc-kit/blob/main/src/generators/jsx-ast/utils/buildContent.mjs#L289-L316, we're building the document tree in JSX here by using this ast.mjs file (https://github.com/nodejs/doc-kit/blob/main/src/generators/jsx-ast/utils/ast.mjs) and this is awkward as hell for me and hard to maintain. If we *ever* want to make change to the document tree or customizing props, this is going to be incredibly hard for newcomers.

I was never fan of this approach but initially it was simple, now it is overly complex. We're hand-crafting JSX elements here, where the reality is we could just make regular React Components (JSX Components/Elements)..

The other part that hinders customization of doc-kit is that we already wire straight on the jsx-ast generator our own ui-components which limit in the future if people want to have other layouts/designs or use their own components for the structure of their docs.

This is more of a long-term ask, but I do feel we should try to make the customization and maintainability here as easy as possible, but the jsx-ast seems to be doing too much stuff and too much hard-coded (https://github.com/nodejs/doc-kit/tree/main/src/generators/jsx-ast)...

It'd be great if the Stability Overview is its own component. Could be on node-core/ui-components or specifically in the components folder on the web generator.

avivkeller:
> The other part that hinders customization of doc-kit is that we already wire straight on the jsx-ast generator our own ui-components which limit in the future if people want to have other layouts/designs or use their own components for the structure of their docs.

That's something that can definitely be changed. For instance, we allow customization of the logo, we can allow customization of everything else.

ovflowd:
> It's a little off putting for new comers, I suppose, however, thats nothing some documentation can't fix.

How you see customization of the tree or document or components with the current hard-coded jsx tree?

ovflowd:
Btw, I still believe this should be a React component. I'm 👎 with the current approach because I know we can make it simpler with React, but if that's too much trouble, I can waive my block.

AugustinMauroy:
@avivkeller agree that I should comme back to react component ?

src/generators/web/ui/components/StabilityOverview/index.jsx
```diff
@@ -0,0 +1,37 @@
+import BadgeGroup from '@node-core/ui-components/Common/BadgeGroup';
+
+const STABILITY_KINDS = ['error', 'warning', 'default', 'info'];
```
ovflowd: nit: move to constants file

src/generators/jsx-ast/utils/buildStabilityOverview.mjs
```diff
@@ -0,0 +1,16 @@
+import { createJSXElement } from './ast.mjs';
+import { JSX_IMPORTS } from '../../web/constants.mjs';
```
ovflowd: @avivkeller knowing we're going to broke down generators in their own things, should the jsx-ast generator use web generator as import?

I'd say that all these constants should reside in the lower denominator (jsx-ast) (like instead of jsx-ast depending on web, web depends on jsx-ast, so it should be the opposite... Otherwise circular dependencies)

src/generators/jsx-ast/utils/buildContent.mjs
```diff
@@ -290,13 +299,16 @@ export const createDocumentLayout = (
   entries,
   sideBarProps,
   metaBarProps,
-  remark
+  remark,
+  stabilityOverviewEntries = []
```
ovflowd: @avivkeller I'm not a fan that this fn is becoming bigger and bigger with sideBarProps, metaBarProps, stabilityOverviewEntries, could we create a share Map for storing these things so we don't need to pass down within fns?

This can be done in a follow-up PR, but I want to keep these methods clean.

src/generators/jsx-ast/utils/buildContent.mjs
```diff
@@ -276,6 +277,14 @@ export const processEntry = (entry, remark) => {
       (parent.children[idx] = createSignatureTable(node, remark))
   );
 
+  // Inject the stability overview table where the slot tag is present
+  if (
```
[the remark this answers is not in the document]
ovflowd: cc @avivkeller I really do wonder if there's an alternative here, because you're kinda always passing the stability overview entries in the hope of finding this, I'd argue it should be done differently/somewhere else?

src/generators/jsx-ast/utils/buildContent.mjs
avivkeller: Agreed, I'm sure we can simplify it somehow.

ovflowd: bump @avivkeller

ovflowd:
cc @AugustinMauroy I think you can rebase already, @avivkeller implemented the requirements, if I recall correctly.

ovflowd:
Bump, @AugustinMauroy

AugustinMauroy:
can someone take hand on this pr I'm too busy

avivkeller:
On it! I'll have this ready in a jiffy!

PR #650, opened 2026-03-06
chore(deps): use package-lock

avivkeller:
Ref: https://github.com/nodejs/node/pull/62139
Ref: https://openjs-foundation.slack.com/archives/CTEDKHCTB/p1772830019698849

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@node-core/doc-kit",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@node-core/doc-kit",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "dependencies": {
         "@actions/core": "^3.0.0",
         "@heroicons/react": "^2.2.0",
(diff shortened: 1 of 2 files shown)


MattIPv4 (commented):
Can we do a review of all our dependencies and check we're comfortable with their semver ranges? This big change here, as you noted, is going to be that usage of this tooling is no longer pinned to the exact dependency versions we define, so we must be comfortable that the semver ranges we define for our dependencies will be safe and aren't going to cause issues for consumers of this tooling.

ovflowd (approved):
I'm stamping with an approval... for the sake of continuity. I won't even check what changed on package-lock as I'm scared.

This is on you Aviv to please verify this won't blow up the universe 😭

ovflowd:
Or maybe yesterday.

MattIPv4 (approved):
I'm also okay with this, but we should be prepared to revert or fix-forward if we run into issues with dependencies not being pinned as we need them.

avivkeller:
I'm going to land this first, since there's not 100% consensus on how pinning should work, and this is blocking some work in nodejs/node.

aduh95:
> @avivkeller added the https://github.com/nodejs/doc-kit/labels/blocked label on Mar 7
>
> @avivkeller commented on Mar 8
>
> I'm going to land this first, since there's not 100% consensus on how pinning should work, and this is blocking some work in nodejs/node.

The timeline here is hard to follow. What was it blocking exactly? My takeaway from the Slack thread is that the issue was not caused by shrinkwrap, was there some other context I'm missing? Also if there was no consensus on how pinning should work, shouldn't that mean we should have stick with shrinkwrap until a consensus has emerged?

PR #654, opened 2026-03-07
chore(deps): change pinning

avivkeller:
cc @ovflowd 

Changes the dependency pinning per our docs. Rolldown was changed to be an exact dependency, as each release candidate may contain breaking changes.


MattIPv4 (approved):
Not sure I totally agree with this approach of exact pinning, given this repo has demonstrated up until this point that it isn't really needed, but happy to approve, given this is what our current policy states. I agree with pinning rolldown specifically though.

AugustinMauroy (approved):

ovflowd (approved):
SGTM although I feel some runtime dependencies could be more stricter.

avivkeller:
I'm going to open a dedicated issue to discuss how/what pinning rules should apply to this repo, since there's not 100% consensus here.

ovflowd:
@avivkeller feel free to rebase, please. Remember the current dependency pinning agreement stands.

bmuenzenmeyer:
did the rebase - think the pinning stands as-is for now.

bmuenzenmeyer:
i'm going to merge so we dont drift again - @avivkeller please cut another issue if you like or havent yet

PR #664, opened 2026-03-09
feat(worker): simplify input passing

avivkeller:
We have no need to pass these inputs twice

### DIFF
--- a/docs/generators.md
+++ b/docs/generators.md
@@ -241,7 +241,7 @@ export async function* generate(input, worker) {
   };
 
   // Stream chunks as they complete
-  for await (const chunkResult of worker.stream(input, input, deps)) {
+  for await (const chunkResult of worker.stream(input, deps)) {
     // Process chunk result if needed
     yield chunkResult;
   }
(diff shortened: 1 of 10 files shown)


avivkeller:
> LGMT ! did this impact perf ?

I doubt it. Inside of `stream` we are really only looking at the length of `items` and the content of `fullInput`, and they are the same object.

avivkeller:
> please check all generator outputs and compare them before after this change with actuall file hash diffs. Want to be extra sure this is not really changing anything :P

It's not. I've checked locally, plus, if there _was_ a change, our GitHub Actions would've left a comment on the PR

ovflowd:
> > please check all generator outputs and compare them before after this change with actuall file hash diffs. Want to be extra sure this is not really changing anything :P
> 
> It's not. I've checked locally, plus, if there _was_ a change, our GitHub Actions would've left a comment on the PR

OBEY 😆

ovflowd:
anyhow, sgtm :D

PR #672, opened 2026-03-13
feat(fs): file system improvements

### LINKED ISSUE #662: Preserve Directory Structure
Right now, doc-kit creates all files at the top-level. We should preserve the directory structure relative to the input. For example, if the input is `docs/**/*.md`, a file in a `docs/my-subdirectory` should be generated at `out/my-subdirectory`.

This is blocking https://github.com/nodejs/nodejs.org/issues/8612

avivkeller:
This PR introduces some file system improvements to resolve #662. Specifically, it:

- Makes a recursive `writeFile` function, allowing us to write files at any depth
- Removes `safeCopy`. This function was meant for running `doc-kit` in parallel, which is **not done,  nor recommended**.
- Adds new `path`  to the metadata entries, allowing us to "see" where an API actually is.

---

**Note:** This slightly changes the output of HTML generators. Instead of relying on a flat index file, they now build the index using page titles to support multiple depths. This means some titles may differ slightly from the index (e.g., `VM` becomes `VM (executing JavaScript)`), but file names themselves remain unchanged. The legacy generator does not implement this functionality, so while its output remains identical under this PR, it does not support recursive output.

Test this output with `node bin/cli.mjs generate -t web -i "../node/doc/{api,}/*.md" -o "out"`

### DIFF
--- a/src/generators/jsx-ast/generate.mjs
+++ b/src/generators/jsx-ast/generate.mjs
@@ -3,7 +3,7 @@ import buildContent from './utils/buildContent.mjs';
 import { getSortedHeadNodes } from './utils/getSortedHeadNodes.mjs';
 import { groupNodesByModule } from '../../utils/generators.mjs';
 import { getRemarkRecma } from '../../utils/remark.mjs';
-import { href } from '../../utils/url.mjs';
+import { relative } from '../../utils/url.mjs';
 
 const remarkRecma = getRemarkRecma();
 
@@ -28,7 +28,7 @@ export async function processChunk(slicedInput, itemIndices, docPages) {
         heading,
         head.path === path
           ? `${head.basename}.html`
-          : `${href(path, head.path)}.html`,
+          : `${relative(path, head.path)}.html`,
       ])
     );
 
--- a/src/generators/metadata/utils/parse.mjs
+++ b/src/generators/metadata/utils/parse.mjs
@@ -21,7 +21,7 @@ import {
 } from './visitors.mjs';
 import { UNIST } from '../../../utils/queries/index.mjs';
 import { getRemark } from '../../../utils/remark.mjs';
-import { href } from '../../../utils/url.mjs';
+import { relative } from '../../../utils/url.mjs';
 import { IGNORE_STABILITY_STEMS } from '../constants.mjs';
 
 // Creates an instance of the Remark processor with GFM support
@@ -63,7 +63,7 @@ export const parseApiDoc = ({ path, tree }, typeMap) => {
 
   // Make all the typeMap links relative to us
   const relativeTypeMap = Object.fromEntries(
-    Object.entries(typeMap).map(([type, url]) => [type, href(url, path)])
+    Object.entries(typeMap).map(([type, url]) => [type, relative(url, path)])
   );
 
   // Handles the normalisation URLs that reference to API doc files with .md extension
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "estree-util-to-js": "^2.0.0",
     "estree-util-visit": "^2.0.0",
     "github-slugger": "^2.0.0",
+    "glob-parent": "^6.0.2",
```
ovflowd: another dep? D:

avivkeller:
> * This should work in a way that you can specify file name formatting then at the very least, or so that the generator (ie: legacy-html and web) decide that path and name of the files.

Generators can still decide the path of the files they want to export, for instance, orama will always export to `orama-db.json`.
   
>   * This means legacy-html keeps telling whatever fs() functions you have to keep creating the same file names

`legacy-html`, and every other generator, are generated the exact same file names, those have not changed.

>   * But this MUST be configurable, because even once we switch to web generator in nodejs/node the links to the files and their # hashes MUST be the same. That is a NON negotiable.

This doesn't affect anchors, or any root level files. It only affects the file names of files at a non-root depth (by outputting them at that same depth)

---

The reason why [the comparator is reporting changes](https://github.com/nodejs/doc-kit/pull/672#issuecomment-4058076819) is because, without ordering from an index file, the sidebar is in a _slightly_ different order:
image

If needed, I can revert that part of the legacy generator.

src/generators/orama-db/generate.mjs
```diff
@@ -50,7 +51,7 @@ export async function generate(input) {
   // Persist
   if (config.output) {
     await writeFile(
-      `${config.output}/orama-db.json`,
+      join(config.output, 'orama-db.json'),
```
avivkeller:
```suggestion
      join(config.output, config.fileName),
```

src/generators/llms-txt/utils/__tests__/buildApiDocLink.test.mjs
```diff
@@ -79,7 +79,7 @@ describe('buildApiDocLink', () => {
   it('handles doc path replacement', () => {
```
avivkeller:
```suggestion
  it('handles custom pageURL template', () => {
```

src/generators/orama-db/generate.mjs
```diff
@@ -1,12 +1,13 @@
 'use strict';
 
-import { writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
 
 import { create, save, insertMultiple } from '@orama/orama';
 
 import { SCHEMA } from './constants.mjs';
```
avivkeller:
```suggestion

```

src/generators/sitemap/generate.mjs
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
-import { readFile, writeFile } from 'node:fs/promises';
+import { readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 
 import { createPageSitemapEntry } from './utils/createPageSitemapEntry.mjs';
 import getConfig from '../../utils/configuration/index.mjs';
+import { writeFile } from '../../utils/file.mjs';
```
avivkeller:
```suggestion
import { writeFile } from '../../utils/file.mjs';
import { populate } from '../../utils/configuration/templates.mjs';
```

src/generators/orama-db/generate.mjs
```diff
@@ -36,7 +37,7 @@ export async function generate(input) {
         description: paragraph
           ? transformNodeToString(paragraph, true)
           : undefined,
-        href: `${entry.api}.html#${entry.heading.data.slug}`,
+        href: `${entry.path.slice(1)}.html#${entry.heading.data.slug}`,
```
avivkeller:
```suggestion
        href: `${entry.api}.html#${entry.heading.data.slug}`,
```

avivkeller:
> Feel free to simplify this and solve for both solutions

It's fairly complete at the moment, WDYT about landing as is

PR #674, opened 2026-03-14
fix(search): remove empty space in search modal

### LINKED ISSUE #670: Empty Space in Search Modal
(I'm not 100% sure if this needs to be resolved here or in `nodejs/nodejs.org`'s `ui-components`)

The search modal contents empty space that should be minimized.

Image

shivxmsharma:
## Description

The `.searchResultsContainer` was missing `min-height: 0`, which caused
`overflow-y: auto` to be ineffective. Flex items default to
`min-height: auto`, meaning the container expanded to show all content
instead of constraining its height and allowing internal scrolling.

**Changes:**
- Replaced `flex-grow: 1` with `flex: 1` (sets `flex-basis` to `0`,
  ensuring space is distributed from zero rather than from content size)
- Added `min-height: 0` to allow the container to shrink below its
  content size, enabling `overflow-y: auto` to work correctly
- Added `margin-top: auto` to `.footer` to keep it pinned at the bottom
  of the modal

## Validation

Open the search modal and type a query that returns many results. The
results list should now scroll within the modal rather than expanding
the modal indefinitely. Footer remains pinned at the bottom on both
mobile and desktop viewports.

## Related Issues

Fixes #670

### Check List

### DIFF
--- a/src/generators/web/ui/components/SearchBox/index.module.css
+++ b/src/generators/web/ui/components/SearchBox/index.module.css
@@ -1,6 +1,6 @@
 .searchResultsContainer {
   display: flex;
-  min-height: 0;
+  flex-grow: 1;
   flex-direction: column;
   overflow-y: auto;
 }


avivkeller:
I'm not 100% sure, like I said in the issue, why don't you try, test, repeat

shivxmsharma:
> I'm not 100% sure, like I said in the issue, why don't you try, test, repeat

Hey @avivkeller, I did more digging and traced the actual root cause — it's not in `index.module.css` at all.

The empty space comes from `.noResultsWrapper` inside `@node-core/ui-components`:

```css
/* packages/ui-components — Search/Results/Empty */
.noResultsWrapper {
  height: 100%;
  padding-top: 2.5rem;
  padding-bottom: 7.75rem; /* ← 124px, this is the empty space */
}
```

That large `padding-bottom` is a vertical centering trick for full-screen mobile but it creates the blank gap on desktop. Any CSS change in `index.module.css` here won't touch it.

I've reverted the changes on this PR since they don't fix the root cause. Going to open a fix directly in `nodejs/nodejs.org` under `packages/ui-components`. Will link it here once it's up.

PR #687, opened 2026-03-17
fix(legacy-json): prevent parseSignature from dropping arguments on u…

ryzrr:
## Description
When the legacy-json generator parses method signatures, it was dropping parameters if the signature was not wrapped in markdown backticks (like `new Console(options)` instead of `` `new Console(options)` ``).

This happens because the parser was blindly using `substring(1, length - 1)` before running its regex, assuming the first and last characters were always backticks or parentheses. This corrupted un-wrapped strings and caused the parameters to be dropped entirely in the generated JSON.

I removed the dangerous substring call and replaced it with a safe backtick replacement regex. This ensures arguments for un-backticked methods and constructors are properly extracted.

## Validation
Run `node --run test` to verify. All tests pass, and I added regression test cases in `parseSignature.test.mjs` to ensure un-backticked signatures are parsed correctly without dropping arguments.

## Related Issues
No related issue, internal codebase discovery.

## Check List
- [x] I have read the Contributing Guidelines and made commit messages that follow the guideline.

### DIFF
--- a/src/generators/legacy-json/utils/__tests__/parseSignature.test.mjs
+++ b/src/generators/legacy-json/utils/__tests__/parseSignature.test.mjs
@@ -273,6 +273,26 @@ describe('parseSignature', () => {
         ],
       },
     },
+    {
+      name: 'handles un-backticked method signatures without dropping arguments',
+      input: {
+        textRaw: 'new Console(options)',
+        markdown: [{ name: 'options' }],
+      },
+      expected: {
+        params: [{ name: 'options' }],
+      },
+    },
+    {
+      name: 'handles standard method signatures without backticks',
+      input: {
+        textRaw: 'foo(a, b)',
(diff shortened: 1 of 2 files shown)


ryzrr:
it's actually not guaranteed to only receive backticks.

`parseSignature` gets its `textRaw` directly from the Markdown AST (e.g, in parseList.mjs via transformNodesToString). If a doc writer formats a heading as plain text (e.g, ### new Console(options)) instead of inline code (e.g., ### \new Console(options)), textRaw arrives without backticks.

and the  previous .substring(1, textRaw.length - 1) blindly chopped the first and last characters regardless, turning new Console(options) into `ew Console(options`

and by using .replace(/^|$/g, '') safely handles both scenarios by ensuring we only strip backticks if they actually exist.

PR #691, opened 2026-03-21
fix: disable wasm highlighter on riscv64

kxxt:
## Description

riscv64 with sv39 has limited virtual memory space, where creating too many (>=20) wasm memory instances fails.
See https://github.com/nodejs/node/pull/60591 for more details.

This PR fixes the following error encountered during `make test-only` on riscv64:

    [07:02:55.797] ERROR: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance
    RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instancemake[1]: *** [Makefile:392: test/addons/.docbuildstamp] Error 1
    make: *** [Makefile:352: test-only] Error 2

I also limited the number of threads to 1 for riscv64 for the same reason, as `@minify-html/wasm` is used in several generators.

While at it, re-enable wasm highlighter for s390x because the corresponding bug has been fixed.

## Validation

Patch `tools/doc/node_modules/@nodejs/doc-kit/src/utils/highlighter.mjs ` and run `make test-only` again on riscv64 linux.

### Check List

### DIFF
--- a/src/utils/configuration/index.mjs
+++ b/src/utils/configuration/index.mjs
@@ -5,6 +5,7 @@ import { coerce } from 'semver';
 
 import { CHANGELOG_URL, populate } from './templates.mjs';
 import { allGenerators } from '../../generators/index.mjs';
+import logger from '../../logger/index.mjs';
 import { parseChangelog, parseIndex } from '../../parsers/markdown.mjs';
 import { enforceArray } from '../array.mjs';
 import { leftHandAssign } from '../generators.mjs';
@@ -33,7 +34,11 @@ export const getDefaultConfig = lazy(() =>
         }),
       },
 
-      threads: cpus().length,
+      // The number of wasm memory instances is severely limited on
+      // riscv64 with sv39. Running multiple generators that use wasm in
+      // parallel could cause failures to allocate new wasm instance.
+      // See also https://github.com/nodejs/node/pull/60591
+      threads: process.arch === 'riscv64' ? 1 : cpus().length,
       chunkSize: 10,
     })
   )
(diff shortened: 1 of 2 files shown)


AugustinMauroy (approved):

ovflowd (approved):

avivkeller (approved):

avivkeller (approved):

joyeecheung:
I opened https://github.com/nodejs/node/issues/62385

joyeecheung:
I opened https://github.com/nodejs/node/pull/62388 to just make the process dependency free (I don't think that block this PR. It just makes PRs like this no longer necessary at all in the future).

avivkeller:
> why do we need a doc-kit + WASM monolith for that when it can just be done with a few regexes?

Once this repo is a monorepo, we won't

kxxt:
A new release on npm would be appreciated because I need to bump it in `tools/doc/package.json` to complete the fix. Thanks!

avivkeller:
I believe a new release has happened since this landed

kxxt:
> I believe a new release has happened since this landed

Thanks! I checked https://github.com/nodejs/doc-kit/commit/3a17176af069d4a446bc87d0ecd8e0113280cd8d and didn't see any tags including that commit. But I forgot to check it on npm myself.

avivkeller:
we removed a WASM guard for S390X as we were under the impression that a V8 patch from a member of @nodejs/platform-s390 resolved the underlying issue.

We can revert that if that's not actually the case.

PR #723, opened 2026-03-27
feat(doc): add specification

### LINKED ISSUE #402: Document structure of parseable documentation markdown
Even though this project's original goal is focussed around parsing the existing Node.js documentation, it would be valuable to start defining the specific ruleset it supports. For example, what is to a Class or Function definition? What goes into the section header, what are the necessary vs option subsections, etc. 

I imagine that once we define the foundational ruleset for the existing Node.js docs, then we can iterate on it and adjust to support a different format and even become more general purpose. This would go hand in hand with linter rules and more specifically what kind of AST structure we want to declare. 

In the past I experimented with this idea. What we are defining is really a subset of markdown. Since markdown itself is not very well defined we should make some declarations upfront like are we starting from GitHub Flavored Markdown or something else? I called my previous experiment "markdown documentation language" or "mddl" for short. You can see my WIP idea [here](https://github.com/Ethan-Arrowood/mddl/tree/main/packages/spec).

It'd be my pleasure to start contributing to this initiative instead of continuing to create my own.

avivkeller:
Closes #402

This PR introduces a formal specification for the Markdown format consumed by doc-kit. I drafted it over the past few days while thinking through how to align webpack's documentation output (for our GSoC) with doc-kit's parseable structure, then revised and formalized it with AI.

The specification documents the Markdown we currently support. However, it defines a `<!--module=node:fs/promises-->` directive, as @Ethan-Arrowood proposed in #402. This allows sub-sections of a document to be associated with a specific module when needed (see §4.4.4 and §6.2.6).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,3 +12,7 @@ out/
 
 # Generated Files
 src/generators/metadata/maps/mdn.json
+
+# The specification uses things that prettier would not
+# approve of, such as bullets with `*`
+docs/specification.md
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1,0 +1,1189 @@
+# Specification
+
+**Version**: 1.0.0
+
+**Authored By**: Aviv Keller (<me@aviv.sh>)
+
+This document specifies the Markdown format consumed by
+[`@node-core/doc-kit`][doc-kit]. It defines the structural, syntactic, and
+semantic rules that documents MUST follow to be correctly parsed. The format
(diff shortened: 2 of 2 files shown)


ovflowd (approved):
SGTM! Really good stuff, this is phenomenal work. I've left a few comments!

AugustinMauroy (approved):
that seem good

Ethan-Arrowood (approved):
This is a great start! Excellent work.

bmuenzenmeyer:
I'm late to this - but this is really stellar!

PR #724, opened 2026-03-27
hotfix(mdn): type map URL fixes

avivkeller:


ovflowd:
@avivkeller even if this is a hot-fix, approvals are required. Not sure if youmerged before I gave my approval lol.

PR #731, opened 2026-03-30
feat(monorepo): implement orchestrator

avivkeller:
## Description

This PR changes our generation logic to, instead of using a generator registry, import generators as specifiers. At the beginning of a pipeline, only the specifiers needed for generation are loaded. Because of this, `generate.mjs` and `index.mjs` don't really even need to be seperate.

## Validation

Every test and generation pipeline should succeed.

### Check List


packages/core/src/generators/legacy-html-all/index.mjs
```diff
@@ -1,28 +1,82 @@
...
+import { join } from 'node:path';
+
+import getConfig from '../../utils/configuration/index.mjs';
+import { minifyHTML } from '../../utils/html-minifier.mjs';
+import { getRemarkRehype } from '../../utils/remark.mjs';
+import { defaultConfiguration as legacyHtmlDefaults } from '../legacy-html/index.mjs';
+import { replaceTemplateValues } from '../legacy-html/utils/replaceTemplateValues.mjs';
+import tableOfContents from '../legacy-html/utils/tableOfContents.mjs';
+
+export const name = 'legacy-html-all';
```
flakey5, line 13: apologizes if this has already been discussed but wouldn't having this export an object still be better for type support at least?

flakey5 (approved):
apologizes for the delay, thank you!

packages/core/src/generators/legacy-html-all/index.mjs
avivkeller: We don't _really_ care about type support for individual generators anymore, since they can be anything, there's no longer a set list. Once we move them to seperate packages, they'll provide their own types

ovflowd (commented):
This PR has so many changes that I'm so against of, that I'm furious this got merged just with one "blank" approval. C'mon @avivkeller you can do better. I said I wanted to review this first 🤦

ovflowd:
(Yes, I'm actually mad at you for merging this huge code change in one go, with one review).

PR #745, opened 2026-04-02
feat: Sidebar enhancements

canerakdas:
## Description

This PR's main goal is to group the links in the sidebar to help developers find the content they're looking for more easily.

As a temporary solution, grouping for the Node.js docs is done using a static array. However, in the long term, we plan to add category information to pages in YAML format and remove this array as well

### Before / After
image
image
<hr />
image
image
<hr />
image
image

## Related Issues

Related to #646, https://github.com/nodejs/learn/issues/6

### Check List


src/generators/web/ui/constants.mjs
ovflowd: Not sure this should belong in this repo.

ovflowd (changes requested):
We shouldn't hard code groups at all. Simply, add the logic for supporting frontmatter, accept that after this gets merged it won't work, but then upstream changes on nodejs/node to then add the proper groups to each file.

ovflowd:
cc @canerakdas seems like it needs rebasing again?

src/generators/web/ui/components/SideBar/utils/index.mjs
```diff
@@ -0,0 +1,104 @@
...
+      value: url.replace('{path}', metadata.path),
+      label,
+    }));
+};
+
+/**
+ * Redirect to a URL
+ * @param {string} url URL
+ */
+export const redirect = url => (window.location.href = url);
```
ovflowd, line 104: I wonder if window ref should be done directly, on the server side this would bork-ish.

ovflowd: Also this seems generic enough it shouldn't really be a fn IDK

src/generators/web/ui/components/SideBar/index.jsx
```diff
@@ -1,67 +1,40 @@
...
- * @returns {number}
- */
-const getMajorVersion = v => parseInt(String(v).match(/\d+/)?.[0] ?? '0', 10);
-
-/**
- * Redirect to a URL
- * @param {string} url URL
- */
-const redirect = url => (window.location.href = url);
+import styles from './index.module.css';
```
ovflowd, line 6: style imports should happen after everything afaik

src/generators/web/ui/types.d.ts
```diff
@@ -30,8 +30,10 @@ declare module '#theme/config' {
     major: number;
   }>;
   export const editURL: string;
-  export const pages: Array<[string, string]>;
-  export const languageDisplayNameMap: Map<string[], string>;
+  export const pages: Array<
+    [number, { heading: string; path: string; category?: string }]
```
ovflowd, line 34: Can `{ heading: string; path: string; category?: string }` be its own type?

src/generators/web/utils/pages.mjs
```diff
@@ -0,0 +1,81 @@
+/**
+ * Default weight used when a page does not define one.
+ */
+export const DEFAULT_PAGE_WEIGHT = -1;
```
ovflowd, line 4: I don't think we should have a page weight system at all. I chatted that with @avivkeller and ordering shouldn't be done via a page weight.

Afaik, what I chatted with @avivkeller is that navigation lists should be manual entries.

ovflowd (commented):
I wonder if this PR is still necessary, now that @avivkeller made a fix already? The fix being that navigation entries are defined on doc-kit config, I think you can sync with @avivkeller what I agreed with him, and maybe if more docs are needed he can fill you in?

But on github.com/nodejs/learn this is at least fixed already?

canerakdas:
> I wonder if this PR is still necessary, now that @avivkeller made a fix already? The fix being that navigation entries are defined on doc-kit config, I think you can sync with @avivkeller what I agreed with him, and maybe if more docs are needed he can fill you in?
> 
> But on github.com/nodejs/learn this is at least fixed already?


I hadn't checked the latest changes in learn 😄 Yeah, it does look unnecessary, I'll go ahead and close this PR for now. 

We can create and add a similar file for the docs later on
https://github.com/nodejs/learn/blob/main/site.json#L111-L385

PR #754, opened 2026-04-06
feat(shiki): more langs

avivkeller:
Adds support for languages used by Node-API (CMake and Python)


ovflowd (approved):

shiki.config.mjs
```diff
@@ -37,6 +39,8 @@ export default {
     ...cLanguage,
     ...cPlusPlusLanguage,
     ...coffeeScriptLanguage,
+    ...cmakeLanguage,
```
ovflowd, line 42: I foresee our bundle doing kaboom

avivkeller: This is server-only is it not?

ovflowd: 🤔 possibly? Hopefully yes?

PR #758, opened 2026-04-09
fix(metadata): preserve -- prefix in anchor IDs for CLI flags

### LINKED ISSUE #757: [v25] Anchors without `--` prefix
### Affected URL(s)

https://nodejs.org/docs/latest-v25.x/api/cli.html#permission

### Description of the problem

- In v24, HTML anchors have the prefix `--`. For example, for the heading `--permission`, the anchor is also [`--permission`](https://nodejs.org/docs/latest-v24.x/api/cli.html#--permission).
- In v25, the prefix is removed. For the heading `--permission`, the anchor is [`permission`](https://nodejs.org/docs/latest-v25.x/api/cli.html#permission).

The links haven't been changed, so they still go to the correct page, but they don't scroll to the right spot.

> - File System - manageable through [--allow-fs-read](https://nodejs.org/docs/latest-v25.x/api/cli.html#--allow-fs-read), [--allow-fs-write](https://nodejs.org/docs/latest-v25.x/api/cli.html#--allow-fs-write) flags

---

Either the problem comes from how the anchors are generated, or all the links need to be corrected.

sujalgoel:
Fixes #757.

In v25, CLI flag headings like `--permission` produce anchor `#permission`
instead of `#--permission`, breaking links that worked in v24.

Two regexes in `DOC_API_SLUGS_REPLACEMENTS` caused this. The leading
hyphen rule (`^-+(?!-*$)`) stripped all leading hyphens unconditionally.
The consecutive hyphen rule (`^(?!-+$).*?(--+)`) also fired on anything
starting with `--`.

Changes in `constants.mjs`:
- Leading hyphen: `^-+(?!-*$)` -> `^-(?=[^-])`. Strips only a single
  `-` not followed by another, so `--permission` passes through.
- Consecutive hyphens: added `[^-]` after the start anchor so the rule
  skips `--`-prefixed slugs.

Changes in `slugger.test.mjs`:
- Updated `--foo` expectation from `'foo'` to `'--foo'`
- Added tests for `--permission` and `--allow-fs-read`

**Test plan**
- [ ] `node --run test`
- [ ] `node --run lint`
- [ ] `node --run format`

### DIFF
--- a/src/generators/metadata/constants.mjs
+++ b/src/generators/metadata/constants.mjs
@@ -5,11 +5,10 @@ export const IGNORE_STABILITY_STEMS = ['documentation'];
 // These are string replacements specific to Node.js API docs for anchor IDs
 export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
-  { from: /&/, to: '-and-' }, // Replace &
-  { from: /[/,:;\\ ]/g, to: '-' }, // Replace /,:;\. and whitespace
-  { from: /^-(?=[^-])/g, to: '' }, // Remove a single leading hyphen (preserves -- prefix for CLI flags)
+  { from: /&/g, to: '-and-', pre: true }, // Replace & before slugging (slugger removes it without a separator)
+  { from: /=/g, to: '-', pre: true }, // Replace = before slugging (slugger removes it without a separator)
+  { from: /[/,:;\\ ]/g, to: '-' }, // Replace /,:;\ and whitespace
   { from: /(?<!^-*)-+$/g, to: '' }, // Remove any trailing hyphens
-  { from: /^(?!-+$)[^-].*?(--+)/g, to: '-' }, // Replace multiple consecutive hyphens (not at start)
 ];
 
 // These are regular expressions used to determine if a given Markdown heading
--- a/src/generators/metadata/constants.mjs
+++ b/src/generators/metadata/constants.mjs
@@ -7,9 +7,9 @@ export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
   { from: /&/, to: '-and-' }, // Replace &
   { from: /[/,:;\\ ]/g, to: '-' }, // Replace /,:;\. and whitespace
-  { from: /^-+(?!-*$)/g, to: '' }, // Remove any leading hyphens
+  { from: /^-(?=[^-])/g, to: '' }, // Remove a single leading hyphen (preserves -- prefix for CLI flags)
   { from: /(?<!^-*)-+$/g, to: '' }, // Remove any trailing hyphens
-  { from: /^(?!-+$).*?(--+)/g, to: '-' }, // Replace multiple hyphens
+  { from: /^(?!-+$)[^-].*?(--+)/g, to: '-' }, // Replace multiple consecutive hyphens (not at start)
 ];
 
 // These are regular expressions used to determine if a given Markdown heading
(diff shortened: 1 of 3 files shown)


src/generators/metadata/utils/__tests__/slugger.test.mjs
```diff
@@ -80,6 +80,16 @@ describe('slug', () => {
...
+  describe('cli flag anchor preservation', () => {
+    it('preserves -- prefix for CLI flags', () => {
+      assert.strictEqual(slug('--permission', identity), '--permission');
+    });
+
+    it('preserves -- prefix for multi-word CLI flags', () => {
+      assert.strictEqual(slug('--allow-fs-read', identity), '--allow-fs-read');
+    });
+  });
+
```
avivkeller: Redundant

avivkeller (approved):
Just make sure existing links aren't broken

src/generators/metadata/utils/__tests__/slugger.test.mjs
sujalgoel: Removed, thanks!

sujalgoel:
> Just make sure existing links aren't broken

Should be fine. The docs never updated their links to match v25's output (the issue even calls this out), so `#--permission` style links were already broken... This puts the anchors back to where they were in v24

regseb:
On this [preview page](https://api-docs-tooling-lmiibhp9n-openjs.vercel.app/cli.html#-conditionscondition), the heading _-C condition, --conditions=condition_ has the anchor `-conditionscondition`. The title is cut off at `--`.

---

You can use [Link Checker](https://validator.w3.org/checklink) to _check links and anchors in Web pages or full Web sites_. For example, with [preview CLI](https://api-docs-tooling-lmiibhp9n-openjs.vercel.app/cli.html):

> Broken fragments:
> - https://api-docs-tooling-lmiibhp9n-openjs.vercel.app/cli.html#-p---print-script (line 609)
> - https://api-docs-tooling-lmiibhp9n-openjs.vercel.app/cli.html#environment-variables_1 (line 344)
> - https://api-docs-tooling-lmiibhp9n-openjs.vercel.app/cli.html#-r---require-module (lines 518, 586)

src/generators/metadata/constants.mjs
```diff
@@ -7,9 +7,9 @@ export const DOC_API_SLUGS_REPLACEMENTS = [
   { from: /node.js/i, to: 'nodejs' }, // Replace Node.js
   { from: /&/, to: '-and-' }, // Replace &
   { from: /[/,:;\\ ]/g, to: '-' }, // Replace /,:;\. and whitespace
-  { from: /^-+(?!-*$)/g, to: '' }, // Remove any leading hyphens
+  { from: /^-(?=[^-])/g, to: '' }, // Remove a single leading hyphen (preserves -- prefix for CLI flags)
   { from: /(?<!^-*)-+$/g, to: '' }, // Remove any trailing hyphens
-  { from: /^(?!-+$).*?(--+)/g, to: '-' }, // Replace multiple hyphens
+  { from: /^(?!-+$)[^-].*?(--+)/g, to: '-' }, // Replace multiple consecutive hyphens (not at start)
```
MattIPv4: This regex looks very odd, unless there is some magic going on, this is replacing all of `[^-].*?(--+)` with `-`?

sujalgoel: > This regex looks very odd, unless there is some magic going on, this is replacing all of `[^-].*?(--+)` with `-`?

Yeah it does! `foo--bar` becomes `-bar` and replaces everything from the start through `--` with a single `-` and it was a pre-existing behavior I kept but I am thinking of removing it in the next commit... regseb's findings show why

sujalgoel:
The broken fragments regseb found (`#-p---print-script`, `#-r---require-module`, `-conditionscondition`) are pre-existing on main and were not introduced here! I believe it's Worth fixing while we're touching this code though. .. github-slugger already produces the right anchors, the transformation rules are what break them. Next push removes steps 4 and 6 and adds `=` to the special chars list

Issue #761, opened 2026-04-10
The title "Stream" in the navigation block is not bold


regseb:
### Affected URL(s)

https://nodejs.org/docs/latest-v25.x/api/stream.html

### Description of the problem

When you're on the [_Stream_](https://nodejs.org/docs/latest-v25.x/api/stream.html) (or [_Modules: `node:module` API_](https://nodejs.org/docs/latest-v25.x/api/module.html)) page, the title of the current page in the left navigation block isn't bold (because it doesn't have the CSS class `active`). For the other pages in the documentation, the title of the current page is bold.

The _Modules: `node:module` API_ page is the only one with code in its title. That might explain its special case. But this explanation doesn’t work for the _Stream_ page.

<img width="1640" height="1060" alt="Image" src="https://github.com/user-attachments/assets/c1e115b7-5c68-4dcf-820b-b1fcc1a90e40" />

moshams272:
On It

PR #762, opened 2026-04-11
fix(legacy-html): prevent partial nav class matches

moshams272:
## Description

The previous `.replace()` logic used a partial match (`class="nav-${head.api}`). For APIs like `stream`, the partial string `class="nav-stream` would incorrectly match the first occurrence it found, which happens to be `class="nav-stream_iter"`. This resulted in corrupted class names `class="active_iter  nav-stream"` and left the actual `stream` link unmodified:

Screenshot from 2026-04-11 12-58-24

The same for  `Modules: node:module API`:

Screenshot from 2026-04-11 12-59-18

**The Fix:**

Added closing double-quotes to the search and replace strings to enforce an exact match (`class="nav-${head.api}"`).

## Validation

- Generated the `legacy-html` docs locally.
- Verified that the `Stream` and `Modules: node:module API` sidebar links are now correctly bolded (`active` class applied).

## Related Issues

#761 

### Check List

### DIFF
--- a/src/generators/legacy-html/generate.mjs
+++ b/src/generators/legacy-html/generate.mjs
@@ -35,8 +35,8 @@ export async function processChunk(slicedInput, itemIndices, navigation) {
     const { head, nodes, headNodes } = slicedInput[idx];
 
     const nav = navigation.replace(
-      `class="nav-${head.api}`,
-      `class="nav-${head.api} active`
+      `class="nav-${head.api}"`,
+      `class="nav-${head.api} active"`
     );
 
     const toc = String(


avivkeller (approved):

bjohansebas (approved):

ovflowd (approved):
Nice, can you verify this doesn't break existing scenarios?

moshams272:
Oh, I forgot to mention that! 🤦🏻‍♂️

Actually, I already verified it. After tracing, I found that the links contain a one class in the attribute (`nav-${api}`) and the only class that added to it is `active`:
Screenshot from 2026-04-12 04-59-55

So, It will replace it correctly, and verified the output locally for every link to ensure that it didn't collision.

PR #766, opened 2026-04-13
fix: resolve flash of unstyled content in dark mode

TusharThakur04:
## Description
This PR resolves the FOUC  by resolving `system` to `dark` or `light` using matchMedia.
Earlier `system` was directly being applied as `set-theme = system`

### before :
[Screencast from 2026-04-13 19-26-34.webm](https://github.com/user-attachments/assets/1b37aadd-fa15-4c04-ba70-5832d318d4b4)

### after : 
[Screencast from 2026-04-13 19-24-03.webm](https://github.com/user-attachments/assets/ba8c5424-b191-40f0-b4e2-37e590f981a9)

## Related Issues
n/a

### Check List

### DIFF
--- a/src/generators/web/template.html
+++ b/src/generators/web/template.html
@@ -19,7 +19,7 @@
     href="https://fonts.googleapis.com/css2?family=IBM+Plex+Mono&family=Open+Sans:ital,wght@0,300..800;1,300..800" />
 
   <!-- Apply theme before paint to avoid Flash of Unstyled Content -->
-  <script>document.documentElement.setAttribute("data-theme", document.documentElement.style.colorScheme = localStorage.getItem("theme") || (matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light"));</script>
+  <script>${themeScript}</script>
   <script type="importmap">${importMap}</script>
   <script type="speculationrules">${speculationRules}</script>
 </head>
--- a/src/generators/web/ui/theme-script.mjs
+++ b/src/generators/web/ui/theme-script.mjs
@@ -1,0 +1,26 @@
+'use strict';
+
+/**
+ * This script is designed to be inlined in the <head> of the HTML template.
+ * It must execute BEFORE the body renders to prevent a Flash of Unstyled Content (FOUC).
+ */
+function initializeTheme() {
+  var THEME_STORAGE_KEY = 'theme';
+  var THEME_DATA_ATTRIBUTE = 'data-theme';
+  var DARK_QUERY = '(prefers-color-scheme: dark)';
+
+  var savedUserPreference = localStorage.getItem(THEME_STORAGE_KEY);
+  var systemSupportsDarkMode = window.matchMedia(DARK_QUERY).matches;
+
+  var shouldApplyDark =
+    savedUserPreference === 'dark' ||
+    (savedUserPreference === 'system' && systemSupportsDarkMode) ||
+    (!savedUserPreference && systemSupportsDarkMode);
+
+  var themeToApply = shouldApplyDark ? 'dark' : 'light';
+
+  document.documentElement.setAttribute(THEME_DATA_ATTRIBUTE, themeToApply);
+  document.documentElement.style.colorScheme = themeToApply;
+}
+
+export var THEME_SCRIPT = `(${initializeTheme.toString()})();`;
--- a/src/generators/web/utils/processing.mjs
+++ b/src/generators/web/utils/processing.mjs
@@ -13,6 +13,7 @@ import getConfig from '../../../utils/configuration/index.mjs';
 import { populate } from '../../../utils/configuration/templates.mjs';
 import { minifyHTML } from '../../../utils/html-minifier.mjs';
 import { SPECULATION_RULES } from '../constants.mjs';
+import { THEME_SCRIPT } from '../ui/theme-script.mjs';
 
 /**
  * Populates a template string by evaluating it as a JavaScript template literal,
@@ -137,6 +138,7 @@ export async function processJSXEntries(entries, template) {
         importMap: clientBundle.importMap?.replaceAll('/', root) ?? '',
         entrypoint: `${data.api}.js?${randomUUID()}`,
         speculationRules: SPECULATION_RULES,
+        themeScript: THEME_SCRIPT,
         root,
         metadata: data,
         config,


src/generators/web/template.html
```diff
@@ -19,7 +19,7 @@
     href="https://fonts.googleapis.com/css2?family=IBM+Plex+Mono&family=Open+Sans:ital,wght@0,300..800;1,300..800" />
 
   <!-- Apply theme before paint to avoid Flash of Unstyled Content -->
-  <script>document.documentElement.setAttribute("data-theme", document.documentElement.style.colorScheme = localStorage.getItem("theme") || (matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light"));</script>
+  <script>(function(){var p=localStorage.getItem("theme");var t=(!p||p==="system")?(matchMedia("(prefers-color-scheme: dark)").matches?"dark":"light"):p;document.documentElement.setAttribute("data-theme",t);document.documentElement.style.colorScheme=t;})();</script>
```
canerakdas: I think we should make this code block more readable for developers who want to contribute or review it. Instead of writing it as a single line, using more descriptive variable names shouldn't significantly impact the file size. @nodejs/web-infra WDYT?

ovflowd: If that's the case this script source should be in its own file and then inlined via a template string {{}}

avivkeller:
This seems highly complex

TusharThakur04:
the `{$themeScript}` literal is being replaced by newly created `themeScript` by `generate.mjs`

it is done so that the script is available before rendering the html and there is no latency which could have happened if browser makes request for the `themeScript.mjs` if we use src inside the script tag

src/generators/web/generate.mjs
```diff
@@ -15,7 +15,15 @@ import { writeFile } from '../../utils/file.mjs';
...
-  const template = await readFile(config.templatePath, 'utf-8');
+  let template = await readFile(config.templatePath, 'utf-8');
+
+  // Read the theme script from UI folder
+  const themeScriptPath = join(import.meta.dirname, 'ui', 'theme-script.mjs');
+  const themeScriptContent = await readFile(themeScriptPath, 'utf-8');
+
+  // Wrap theme script in script tag and replace the placeholder
+  const inlinedThemeScript = `<script>${themeScriptContent}</script>`;
+  template = template.replace('${themeScript}', inlinedThemeScript);
```
ovflowd: Don't use .replace(), I believe that @avivkeller implemented a system to automatically evaluate such template variables?

src/generators/web/template.html
```diff
@@ -18,8 +18,12 @@
...
     href="https://fonts.googleapis.com/css2?family=IBM+Plex+Mono&family=Open+Sans:ital,wght@0,300..800;1,300..800" />
 
-  <!-- Apply theme before paint to avoid Flash of Unstyled Content -->
-  <script>document.documentElement.setAttribute("data-theme", document.documentElement.style.colorScheme = localStorage.getItem("theme") || (matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light"));</script>
+<!--
+  This placeholder is dynamically replaced by generator.mjs.
+  It injects an inline script that initializes the theme (light/dark)
+  before the page renders to avoid Flash of Unstyled Content (FOUC).
+-->
+  ${themeScript}
```
ovflowd:
```suggestion
  <script type="javascript">${themeScript}</script>
```

src/generators/web/template.html
```diff
@@ -18,8 +18,12 @@
   <link rel="stylesheet"
     href="https://fonts.googleapis.com/css2?family=IBM+Plex+Mono&family=Open+Sans:ital,wght@0,300..800;1,300..800" />
 
-  <!-- Apply theme before paint to avoid Flash of Unstyled Content -->
-  <script>document.documentElement.setAttribute("data-theme", document.documentElement.style.colorScheme = localStorage.getItem("theme") || (matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light"));</script>
+<!--
```
ovflowd: No need for this comment.

src/generators/web/generate.mjs
```diff
@@ -15,7 +15,15 @@ import { writeFile } from '../../utils/file.mjs';
 export async function generate(input) {
   const config = getConfig('web');
 
-  const template = await readFile(config.templatePath, 'utf-8');
+  let template = await readFile(config.templatePath, 'utf-8');
+
+  // Read the theme script from UI folder
+  const themeScriptPath = join(import.meta.dirname, 'ui', 'theme-script.mjs');
+  const themeScriptContent = await readFile(themeScriptPath, 'utf-8');
```
ovflowd: We shouldn't ue `fs` the idea is to use our rolldown pipeline for this script, so it gets properly minified and transformed.

ovflowd: Otherwise the file doesn't get minified, which is the idea behind separating src<>final result.

TusharThakur04:
@ovflowd 
please review, i ve made the changes you suggested

src/generators/web/ui/theme-script.mjs
```diff
@@ -0,0 +1,25 @@
...
+  const shouldApplyDark =
+    savedUserPreference === 'dark' ||
+    (savedUserPreference === 'system' && systemSupportsDarkMode) ||
+    (!savedUserPreference && systemSupportsDarkMode);
+
+  const themeToApply = shouldApplyDark ? 'dark' : 'light';
+
+  document.documentElement.setAttribute(THEME_DATA_ATTRIBUTE, themeToApply);
+  document.documentElement.style.colorScheme = themeToApply;
+})();`.trim();
```
avivkeller: If we _do_ go with the "custom script file" way, we should import with fs and not a string

TusharThakur04: https://github.com/nodejs/doc-kit/pull/766#discussion_r3092507112

canerakdas: Instead of turning the entire script into a string, converting the function using `.toString()` makes the code more readable in my opinion;

```mjs
'use strict';

/**
 * This script is designed to be inlined in the <head> of the HTML template.
 * It must execute BEFORE the body renders to prevent a Flash of Unstyled Content (FOUC).
 */
function initializeTheme() {
  const THEME_STORAGE_KEY = 'theme';
  const THEME_DATA_ATTRIBUTE = 'data-theme';
  const DARK_QUERY = '(prefers-color-scheme: dark)';

  const savedUserPreference = localStorage.getItem(THEME_STORAGE_KEY);
  const systemSupportsDarkMode = window.matchMedia(DARK_QUERY).matches;

  const shouldApplyDark =
    savedUserPreference === 'dark' ||
    (savedUserPreference === 'system' && systemSupportsDarkMode) ||
    (!savedUserPreference && systemSupportsDarkMode);

  const themeToApply = shouldApplyDark ? 'dark' : 'light';

  document.documentElement.setAttribute(THEME_DATA_ATTRIBUTE, themeToApply);
  document.documentElement.style.colorScheme = themeToApply;
}

export const THEME_SCRIPT = `(${initializeTheme.toString()})();`;
```

src/generators/web/ui/theme-script.mjs
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+/**
+ * This script is designed to be inlined in the <head> of the HTML template.
+ * It must execute BEFORE the body renders to prevent a Flash of Unstyled Content (FOUC).
+ */
+function initializeTheme() {
+  const THEME_STORAGE_KEY = 'theme';
```
canerakdas: Even though `const` is [widely available](https://caniuse.com/const), since this scripts will run directly in the browser, it’s safer to keep it as `var`

canerakdas (approved):
LGTM, I just left one comment 🚀

avivkeller (approved):

PR #767, opened 2026-04-13
fix(cli): accept URL value for --git-ref option

Sahilshrma31:
## Description

Fixes a CLI argument parsing issue where the `--git-ref` option was incorrectly defined as a flag instead of accepting a value. 

The option is intended to take a URL (as indicated by its description), but its previous definition caused the CLI to throw an error when a value was passed. This change updates the option to correctly accept a value.

### Before:
Screenshot 2026-04-14 at 12 19 22 AM

Running the CLI with a value for `--git-ref` fails because the option is incorrectly parsed as a boolean flag instead of accepting a value.

## After:
Screenshot 2026-04-14 at 12 21 26 AM

The `--git-ref` option is now correctly parsed as a value option, allowing a URL to be passed without causing argument parsing errors.

## Related Issues

N/A

### Check List


avivkeller (approved):

avivkeller (changes requested):
I don't think it's supposed to take a URL

Sahilshrma31:
> I don't think it's supposed to take a URL

thanks for the feedback. my reasoning was based on consistency with nearby options like --changelog <url>, --index <url>, and --type-map <url>, and their descriptions mentioning “url or path.” seeing --git-ref next to those made me interpret it similarly, so i updated it to accept <url>

PR #770, opened 2026-04-14
fix(utils): handle raw hashes and queries in url relative() + add tests

ryzrr:
Description
This PR adds the missing unit tests for the relative utility in `src/utils/url.mjs` and patches a minor logic flaw in how it handles hashes and query strings. I included a quick short-circuit to ensure that raw fragments (like `#fragment`) stay completely intact instead of being mistakenly processed as directory paths.

**Note: This is not actively breaking the site right now because no component
currently passes a raw hash to the resolver. However, it is a dormant footgun
that could easily cause unexpected bugs if routing or navigation changes in the future.***

Validation
I ran the Node test suite locally to verify the new url.test.mjs file. The tests confirm that all URL resolution edge cases are handled perfectly and that the existing logic remains completely untouched.

Related Issues
No Related Issues.

Check List
- [x] I have read the Contributing Guidelines and made commit messages that follow the guideline.

### DIFF
--- a/src/utils/__tests__/url.test.mjs
+++ b/src/utils/__tests__/url.test.mjs
@@ -1,0 +1,44 @@
+import { strict as assert } from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { relative } from '../url.mjs';
+
+describe('url util', () => {
+  describe('relative()', () => {
+    it('returns the same URL if it includes a protocol', () => {
+      assert.equal(
+        relative('https://example.com', '/foo/bar'),
+        'https://example.com'
+      );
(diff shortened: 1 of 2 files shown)


avivkeller (changes requested):
Why is this needed?

ryzrr:
I was just trying to add some missing unit test for the `relative` utility to get better coverage. But while writing ,  I noticed the current logic treats raw hashes and query strings as like directory paths.

It isn't breaking anything right now, so it's not urgent. I just figured it made sense to fix it while I was  in the file adding the tests.

PR #771, opened 2026-04-15
fix(web): add noopener and noreferrer to external source links

Sahilshrma31:
## Description

Adds rel="noopener noreferrer" to source-code links generated by the web docs pipeline when links open in a new tab (target="_blank").

The source link rendering in jsx-ast currently opens external GitHub links in a new tab but does not set a rel attribute. This change hardens link behavior by preventing opener access and aligning with standard security best practices for external links.

## Validation

I validated this by generating docs and inspecting rendered source-code links in the output HTML.

### Before
Generated source links had:

target="_blank"
no rel attribute

### After
Generated source links now have:

target="_blank"
rel="noopener noreferrer"

This preserves existing behavior (open in new tab) while adding safer link semantics.

## Related Issues
N/A

### Check List

### DIFF
--- a/src/generators/jsx-ast/utils/buildContent.mjs
+++ b/src/generators/jsx-ast/utils/buildContent.mjs
@@ -90,6 +90,7 @@ export const createSourceLink = sourceLink => {
           {
             href: `${populate(GITHUB_BLOB_URL, config)}${sourceLink}`,
             target: '_blank',
+            rel: 'noopener noreferrer',
           },
           [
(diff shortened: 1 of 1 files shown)


araujogui (approved):

bmuenzenmeyer (approved):
this is probably overkill for a GH link (we trust them more implicitly) but no harm here. noopener is redundant in most browsers, but again, not much harm

((we do both within nodejs.org anyways))

PR #778, opened 2026-04-23
chore(monorepo): break out generators

avivkeller:
This PR breaks out the monorepo feature branch into:

- `@doc-kittens/legacy` (`legacy-*`)
- `@doc-kittens/react` (`orama-db`, `web` TBR `react-html`, `jsx-ast`)
- `@doc-kittens/web` (`sitemap`, `llms-txt`)
- `@doc-kittens/internal` (`ast-*`, `metadata`)
- `@doc-kittens/extras` (Everything else)
-


ovflowd:
@avivkeller can you disable automatic copilot reviews in your acc? 👀

ovflowd:
> This PR breaks out the monorepo feature branch into:
> 
> * `@doc-kittens/legacy` (`legacy-*`)

Can we call it @node-core/doc-kit-legacy-html, @node-core/doc-kit-legacy-json, since eventually they will be removed and they are node-core only.

> * `@doc-kittens/react` (`orama-db`, `web`, `jsx-ast`)

I think this must be split. IMO jsx-ast should be part of the core of doc-kit, not in a separate package. orama-db should be @doc-kittens/orama-db... I don't know what the web one should be since it is right now Node.js-specific by default and probably in the future should use neutral colours and a neutral logo by default without any Node.js-specific theming. And if this is the react one, it should probably be called @doc-kittens/react.

> * `@doc-kittens/web` (`sitemap`, `llms-txt`)

Should be @doc-kittens/sitemap, @doc-kittens/llms

> * `@doc-kittens/internal` (`ast-*`, `metadata`)

IMO, these should be shipped with doc-kit, not in an extra package.

> * `@doc-kittens/extras` (Everything else)

What would be everything else? Examples?

avivkeller:
The idea behind the monorepo was to avoid generators importing packages and files they don't actually need.

For legacy-html, for instance, it depends on some utilities in legacy-json, so it makes sense to put them in the same package.

The extras package isn't a permanent name, it's merely holding everything not already in a package for now. I'll probably break those out into individual packages.

avivkeller:
> IMO, these should be shipped with doc-kit, not in an extra package.


I disagree, in the event that a generator doesn't want that data, and merely wants to create its own AST generator and whatnot, there's no need for it to require these

avivkeller:
> Can we call it @node-core/doc-kit-legacy-html, @node-core/doc-kit-legacy-json, since eventually they will be removed and they are node-core only.

-1, the current name is IMO fine. The scope doesn't really matter.

ovflowd:
> > Can we call it @node-core/doc-kit-legacy-html, @node-core/doc-kit-legacy-json, since eventually they will be removed and they are node-core only.
> 
> -1, the current name is IMO fine. The scope doesn't really matter.

-1 on your -1, the scope does matter.

ovflowd:
> > IMO, these should be shipped with doc-kit, not in an extra package.
> 
> I disagree, in the event that a generator doesn't want that data, and merely wants to create its own AST generator and whatnot, there's no need for it to require these

The core generators should be part of doc-kit, it doesn't make sense that even the basic ones such as ast, metadata, jsx-ast, etc are in individual packages, they should be shipped with doc-kit, and not be even more packages.

ovflowd:
> The idea behind the monorepo was to avoid generators importing packages and files they don't actually need.
> 
> For legacy-html, for instance, it depends on some utilities in legacy-json, so it makes sense to put them in the same package.
> 

Then said utilities should not be shared between them and be part of doc-kit itself. Except the legacy ones, all legacy generators should be their own package.

> The extras package isn't a permanent name, it's merely holding everything not already in a package for now. I'll probably break those out into individual packages.

Can you still please describe what's inside extra? It doesn't make sense at all having a package called "extra" how are people going to know what's inside extra? The idea is that each of the non-core generators is their own isolated package, so that users can install the generators they want. These packages shouldn't be a set of generators. Think of how Radix UI does this.

ovflowd (dismissed):
Formally blocking the PR.

ovflowd:
Forgot to post this before:

Also to clarify, I just want to understand the process of thought, because I thought that the idea I presented originally was quite clear and thought the execution was also clear 😅

avivkeller:
@nodejs/web-infra PTAL

avivkeller:
Bump again @nodejs/web-infra @flakey5

flakey5 (commented):
the ai comments look to be relevant, also why break out the generators' utils away from the generator and into their own directories?

avivkeller:
> the ai comments look to be relevant, also why break out the generators' utils away from the generator and into their own directories?

What do you mean?

avivkeller:
Adding to the agenda since, while we discussed attempting to reduce dependency sizes without the need for a 'break-out', we haven't discussed other aspects of a breakout, for instance, allowing man page linting without OpenSSL

avivkeller:
@nodejs/web-infra Can we move forward with this? Obviously, things have changed and I'd need to rework this PR and it's feature branch, but are we all +1 on monorepo-ifying?

avivkeller:
Bump!

bmuenzenmeyer:
> @nodejs/web-infra Can we move forward with this? Obviously, things have changed and I'd need to rework this PR and it's feature branch, but are we all +1 on monorepo-ifying?

@avivkeller we are still +1 - if there are major concerns to work out, we should align on them before you spend time coding

avivkeller:
I need to open a new PR, this feature branch is outdated. I'll begin working on that :-)

PR #798, opened 2026-05-17
fix(web): shorten method labels in sidebar

panva:
The TOCs are barely usable when methods include their possibly many parameters.

<table>
<tr>
 <td>before
 <td>after
<tr>
 <td>
sidebar-before-webcrypto-toc

 <td>
sidebar-after-webcrypto-toc

</table>

### DIFF
--- a/src/generators/jsx-ast/utils/__tests__/buildBarProps.test.mjs
+++ b/src/generators/jsx-ast/utils/__tests__/buildBarProps.test.mjs
@@ -83,4 +83,71 @@ describe('extractHeadings', () => {
     const result = extractHeadings(entries);
     assert.equal(result.length, 0);
   });
+
+  it('labels callables with the bare name instead of the full signature', () => {
+    const entries = [
+      {
+        heading: {
+          depth: 2,
+          data: {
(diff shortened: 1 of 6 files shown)


avivkeller (changes requested):
These are intentionally expanded, since the short version credits scenarios were multiple headings have the same navigation entry.

The correct approach to shortening these would be to revert the change that expanded them, and consolidate duplicate ToC entries into one.

panva:
> These are intentionally expanded

Bold.

There are only ~26-29 collisions in all our docs. I've updated it to keep the compact form when it is unique otherwise keep the full one.

Screenshot 2026-05-17 at 20 39 39

avivkeller:
cc @nodejs/web-infra I believe the intentional expansion is the method of choice, as it doesn't risk losing any important information, wdyt?

panva:
Ping. The current proposal keeps the unique entries short and only expands when there would be conflicts. There should be no question about the improvement to legibility, clarity, and usability.

avivkeller (approved):

avivkeller (changes requested):
give me a moment to make a change

avivkeller (approved):
give me a moment to make a sense

PR #808, opened 2026-05-22
fix: remove redundant TOC picker from legacy HTML template.

ubay1:
Removed the `__TOC_PICKER__` placeholder from the template and its corresponding generation logic in `replaceTemplateValues.mjs` to avoid rendering a duplicate Table of Contents in the API documentation layout.

## Description

This change removes the extra Table of Contents picker rendered in the content header area of the generated API documentation pages.

The sidebar navigation already provides the full documentation index, so rendering another TOC picker in the page header created redundant navigation and unnecessary visual clutter.

This update removes:

* The `__TOC_PICKER__` placeholder from the HTML template
* The related generation logic in `replaceTemplateValues.mjs`

## Validation

* Generated API docs locally using `doc-kit`
* Verified that the duplicate TOC section no longer appears in the content header
* Confirmed sidebar navigation and other pickers continue to function normally

## Related Issues

Fixes https://github.com/nodejs/node/issues/60667
### Check List

* [x] I have read the [[Contributing Guidelines](https://github.com/nodejs/api-docs-tooling/blob/main/CONTRIBUTING.md)](https://github.com/nodejs/api-docs-tooling/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
* [x] I have run `node --run test` and all tests passed.
* [x] I have check code formatting with `node --run format` & `node --run lint`.
* [ ] I've covered new added functionality with unit tests if necessary.Removed the `__TOC_PICKER__` placeholder from the template and its corresponding generation logic in `replaceTemplateValues.mjs` to avoid rendering a duplicate Table of Contents in the API documentation layout.

### DIFF
--- a/bin/commands/generate.mjs
+++ b/bin/commands/generate.mjs
@@ -56,6 +56,12 @@ export default new Command('generate')
   .addOption(new Option('-c, --changelog <url>', 'Changelog URL or path'))
   .addOption(new Option('--git-ref', 'Git ref URL'))
   .addOption(new Option('--index <url>', 'index.md URL or path'))
+  .addOption(
+    new Option(
+      '--index-redirect-url <url>',
+      'URL to redirect index.html to (legacy-html generator)'
+    )
+  )
   .addOption(new Option('--minify', 'Minify?'))
   .addOption(new Option('--type-map <url>', 'Type map URL or path'))
 
--- a/src/generators/legacy-html/index.mjs
+++ b/src/generators/legacy-html/index.mjs
@@ -31,6 +31,7 @@ export default createLazyGenerator({
     ref: 'main',
     pageURL: '{baseURL}/latest-{version}/api{path}.html',
     editURL: `${GITHUB_EDIT_URL}/doc/api{path}.md`,
+    indexRedirectURL: '',
   },
 
   hasParallelProcessor: true,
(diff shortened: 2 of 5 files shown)


avivkeller (changes requested):
This affects every page of an already legacy UI, why?

ubay1:
> This affects every page of an already legacy UI, why?

Sorry, I misunderstood the scope of the issue initially. You're right that the original approach was too broad since it affected all pages in the legacy UI.

I've reworked the implementation so the behavior is now opt-in via a new `--index-redirect-url` configuration. This allows consumers (like `nodejs/node`) to redirect only the generated `index.html` landing page to something like `documentation.html` without changing the behavior of other legacy pages by default.

ubay1:
Screen Recording 2026-05-23 at 04 04 04

avivkeller:
I don't see the benefit of this, I only see the added complexity

ubay1:
> I don't see the benefit of this, I only see the added complexity

The goal is to support consumers that want `documentation.html` to be the primary entry point for their documentation.

Currently, opening the root URL always leads to the generated TOC page (`index.html`), which only contains a list of links. Some consumers may prefer to take users directly to the documentation landing page instead.

This change is optional and only applies when an `indexRedirectUrl` is explicitly configured. The default behavior remains unchanged.

For example, this is needed by nodejs.org because we don't want to expose the generated TOC page as the documentation entry point and would rather direct users to the documentation landing page.

PR #814, opened 2026-05-30
fix(queries): allow complex types in type linking regex

moshams272:
## Description

Currently, `doc-kit` fails to parse and link complex JSDoc types that contain generics, unions and arrow functions. 

This happens because the regex in `src/utils/queries/index.mjs` uses `[^<({})>]+`, which excludes `<`, `>`, `(`, and `)`. When the parser encounters these characters, it ignores the entire type block, leaving it unlinked and sometimes causing Markdown parsers (Remark) to see them as HTML tags.

This PR updates the regex exclusion list to `[^{}]+`.

## Validation

**Before**:
Screenshot from 2026-05-30 15-13-38

**After**:
Screenshot from 2026-05-30 15-19-35

## Related Issues

No related one.

### Check List

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@node-core/doc-kit",
-  "version": "1.3.6",
+  "version": "1.3.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@node-core/doc-kit",
-      "version": "1.3.6",
+      "version": "1.3.7",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


avivkeller (approved):
Can you run `npm version patch`?

AugustinMauroy (approved):
LGMT ! is there are test for that ?

moshams272:
> LGMT ! is there are test for that ?

I initially relied only on screenshots to show it works, but u r right.

Issue #815, opened 2026-05-30
Reduce memory footprint


avivkeller:
Ref: https://github.com/nodejs/node/pull/62045

PR #816, opened 2026-05-30
chore: reduce memory footprint

avivkeller:
Reduces the memory used by doc-kit by evicting cache after it is no longer needed, and avoiding cloning objects

Ref: #815


avivkeller:
cc @nodejs/web-infra

avivkeller:
Bump @nodejs/web-infra

flakey5 (approved):

PR #841, opened 2026-06-16
fix(legacy-json): preserve source subdirectories in output paths

### LINKED ISSUE #838: Legacy JSON outputting to the wrong directory when subdirectories involved
As titled. `legacy-json` is outputting in the format `generated-slug.json`, not `path/to/file.json`

shivxmsharma:
## Description

Right now `legacy-json` writes every file straight into the output root and names it after the generated `api` slug, so a doc like `path/to/file.md` ends up as `path-to-file.json` instead of `path/to/file.json`. The folder structure of the source docs gets flattened.

There were two reasons for this in `src/generators/legacy-json/generate.mjs`:

1. The output filename was built from `section.api`. That value is a slug (created in `metadata/utils/parse.mjs`) that deliberately turns path separators into dashes, which is exactly why the subfolders disappear. The real relative path is already available on the head node as `head.path` ( the same field `legacy-html` uses ) so I now build a small `api → path` lookup from the head nodes and use that for the filename.

2. The file was written with `writeFile` from `node:fs/promises`, which doesn't create missing folders, so writing into a subfolder would fail even with the right path. I switched to the project's own `writeFile` from `utils/file.mjs`, which creates the parent directories first (again, the helper `legacy-html` already uses).

The change is contained to the write path: `processChunk` and the yielded sections are untouched, so nothing downstream (like `legacy-json-all`) is affected.

## Validation

Running the generator on a few docs that live in subfolders:

```bash
node bin/cli.mjs generate -i "docs/**/*.md" -t legacy-json -o out -v 18.0.0
```

Before:

```
out/alpha.json
out/sub-beta.json
out/sub-deepgamma.json
```

After:

```
out/alpha.json
out/sub/beta.json
out/sub/deep/gamma.json
```

The subfolders are preserved and created automatically, and the generated JSON is unchanged otherwise.

If you're reviewing, the thing to check is that the output path now comes from `path` (via `withExt(path, 'json')`) instead of `section.api`, and that `writeFile` is imported from `utils/file.mjs`.

`node --run test` passes for me. The only suites that fail (`api-links`, `jsx-ast`, `web`) fail the exact same way on a clean `main`, so they're unrelated to this change. `node --run lint` and `node --run format` are clean.

One related thing I noticed but left out on purpose: `buildSection.mjs` builds its `source` field as `doc/api/${head.api}.md` (you can see it in the output above as `doc/api/sub-deepgamma.md`), which has the same slug-vs-path problem for files in subfolders. I kept this PR focused on the output path since that's what the issue is about. Happy to fix `source` here too or in a follow-up, whichever you'd prefer.

## Related Issues

Fixes #838

### Check List

- [x] I have read the [[Contributing Guidelines](https://github.com/nodejs/doc-kit/blob/main/CONTRIBUTING.md)](https://github.com/nodejs/doc-kit/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.


AugustinMauroy (approved):

avivkeller (approved):

avivkeller:
@shivxmsharma can you run `npm version patch`?

shivxmsharma:
> @shivxmsharma can you run `npm version patch`?

Done👍

PR #842, opened 2026-06-16
chore(memory): improve

avivkeller:
After this change, during testing, memory didn't exceed ~1.15GB (~630MB without `all` generation) (before this change, it was 2GB and climbing).

---

Performs AST -> JS on the `jsx-ast` worker side, rather than the `web` main thread, resulting in less information passed between threads.


avivkeller:
Bump @nodejs/web-infra

flakey5 (approved):

PR #883, opened 2026-07-05
fix(metadata): support TS prefix operators in typeParser

moshams272:
## Description

Currently, the `typeParser` in `src/generators/metadata/utils/typeParser.mjs` fails to resolve types that have prefix operators such as `typeof`, `keyof`, `readonly`, and `unique`.

Because the parser treats the prefix and the target type as a single continuous string (e.g., `"typeof Compiler"`), it fails to find a match in the `type-map`, resulting in dead links and plain text rendering in the documentation.

This PR introduces a check for these common prefix operators before falling back to array or generic processing. It strips the prefix, formats it as plain text, and recursively passes the remaining target type to `resolveOr` so it can be correctly matched and linked.

## Validation

Tested it on `webpack-doc-kit`:

**Before:**
image

**After:**
image

## Related Issues

No related issues.

### Check List

### DIFF
--- a/src/generators/metadata/utils/typeParser.mjs
+++ b/src/generators/metadata/utils/typeParser.mjs
@@ -206,6 +206,14 @@ export const parseType = (typeString, transformType) => {
     return parts.map(p => resolveOr(p, transformType)).join(joiner);
   }
 
+  const PREFIXES = ['typeof ', 'keyof ', 'readonly ', 'unique '];
+  for (const prefix of PREFIXES) {
+    if (trimmed.startsWith(prefix)) {
+      const rest = trimmed.slice(prefix.length).trim();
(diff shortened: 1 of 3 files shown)


src/generators/metadata/utils/typeParser.mjs
```diff
@@ -206,6 +206,14 @@ export const parseType = (typeString, transformType) => {
     return parts.map(p => resolveOr(p, transformType)).join(joiner);
   }
 
+  const PREFIXES = ['typeof ', 'keyof ', 'readonly ', 'unique '];
```
avivkeller: This should go at the top level

avivkeller (approved):

AugustinMauroy (approved):

PR #887, opened 2026-07-08
feat(type-parsing): use `oxc-parser`

### LINKED ISSUE #839: `typeParser.mjs` should place functions within code blocks.
Currently, functions transformed into types are transformed into prose. They should instead be transformed into some variation of:

```
`<code>&lt;${sig} =&gt; ${resolveOr(right, transformType)}\`&gt;</code>`
```

avivkeller:
Screenshot 2026-07-08 at 2 20 39 PM
Screenshot 2026-07-08 at 2 21 14 PM


avivkeller:
The diff [was too big to leave as a comment](https://github.com/nodejs/doc-kit/actions/runs/28976525192/job/85984947214)

docs/specification.md
bmuenzenmeyer, line 1: This changes the version of the spec right? How have we been handling this?

package.json
```diff
@@ -87,6 +87,7 @@
     "semver": "^7.8.5",
     "shiki": "^4.3.0",
     "tinyglobby": "^0.2.17",
+    "typescript": "^6.0.3",
```
bmuenzenmeyer: TS7 just came out ([yesterday](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/)), boasting at times an order of magnitude faster experience.

bmuenzenmeyer (commented):
a first pass

package.json
AugustinMauroy: ts7 didn't have js api yet

docs/specification.md
avivkeller: This would be the first change to the spec, however, it's more of a change in wording that behavior. The only behavioral change is that `{string}|{boolean}` must now be `{string|boolean}`. Previously, the spec didn't explicitly say whether to do this or not to do this, now it explicitly says to keep all types with the `{...}` brackets.

avivkeller: 1.0.0 -> 1.1.0 sounds good?

bmuenzenmeyer: i am less concerned if we think its still valid 1.0.0 due to omission - but we should develop some governance over what constitutes a change. imagine semver would work in moooooooost cases

avivkeller:
@bmuenzenmeyer i ended up patching instead of minoring since we previously claim to support these types and similar ones, and this is more of an improvement on that claim than a new feature of the spec.

avivkeller:
Bump @nodejs/web-infra

src/utils/type-annotations/__tests__/hast.test.mjs
```diff
@@ -0,0 +1,117 @@
...
+        { start: 0, end: 9, text: 'vm.Module', href: 'vm.html#class-module' },
+      ],
+    });
+
+    const html = toHtml(typeAnnotationToHighlightedHast(state, node));
+
+    // Shiki tokenizes `vm`, `.`, `Module` separately; the whole range links
+    const anchors = html.match(/href="vm.html#class-module"/g);
+    assert.ok(anchors.length >= 1);
+    assert.match(html.replace(/<[^>]+>/g, ''), /^vm\.Module$/);
```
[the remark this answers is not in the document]
AugustinMauroy: we don't care we are on test and this code will took more computing time so we didn't have any interest of doing that

AugustinMauroy:
do you know what the impact in terms:
- of doc-kit npm size ? 
- since it's a fully parsing job it's ask more computing so what happened to perf 

Is it possible to achieve this with other parser such as oxc/swc which is lighter that full typescript package 

somethign like that:
```ts
import { parseSync, Visitor } from "oxc-parser";

function parseTypeValues(values: string[]) {
  // Same trick as today: wrap each type in a synthetic alias.
  const source = values
    .map((value, i) => `type $T${i} = ${value};`)
    .join("\n");

  const result = parseSync("types.ts", source);

  if (result.errors.length > 0) {
    // Same fallback strategy as today:
    // parse each type individually to isolate invalid annotations.
    return fallback(values);
  }

  const identifiersPerAlias = values.map(() => []);

  const visitor = new Visitor({
    TSTypeReference(node) {
      // Promise<T> / vm.Module / Foo.Bar
      record(node.typeName);
    },

    TSTypeQuery(node) {
      // typeof Foo
      record(node.exprName);
    },

    TSImportType(node) {
      // import("fs").Stats
      record(node.qualifier);
    },

    TSThisType(node) {
      record(node);
    },

    TSLiteralType(node) {
      // null
      if (node.literal.type === "NullLiteral") {
        record(node);
      }
    },

    TSStringKeyword: record,
    TSNumberKeyword: record,
    TSBooleanKeyword: record,
    TSAnyKeyword: record,
    TSUnknownKeyword: record,
    TSNeverKeyword: record,
    TSVoidKeyword: record,
    TSUndefinedKeyword: record,
    TSObjectKeyword: record,
    TSBigIntKeyword: record,
    TSSymbolKeyword: record,
  });

  visitor.visit(result.program);

  return identifiersPerAlias;

  function record(node) {
    // node.range / node.span provides source offsets
    // determine which synthetic alias this belongs to
    // convert to offsets relative to the original annotation
    // push { start, end, text, lookup }
  }
}
```

avivkeller:
@AugustinMauroy nice suggestion! Moved to oxc

src/generators/web/ui/index.css
```diff
@@ -117,4 +117,20 @@ main {
...
+    .type-link {
+      /* Links keep their token color; underline marks the affordance */
+      color: inherit;
+      text-decoration: underline dotted;
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+  }
```
AugustinMauroy, line 135: why do we need that here ? why not on ui component ?

AugustinMauroy:
@avivkeller pleasure to help. 

other part to think if I recall correctly the current import use node native addon and their are a wasm target. Take a look for node core support with our 2000 different distro in CI

avivkeller:
Bump @bmuenzenmeyer

avivkeller:
@AugustinMauroy if this fails CI, I'll switch it out, but let's have hope for you :-)

avivkeller:
This is blocking webpack, so I'd love to land this when we can

bmuenzenmeyer (approved):

src/generators/web/ui/index.css
avivkeller: Cause it's not a UI-component :-), it's a styled span

PR #893, opened 2026-07-11
web: don't render the search bar when orama-db isn't enabled

### LINKED ISSUE #891: `web` generator should not render the search bar if `orama-db` is not enabled

devthedevil:
## Summary
The `web` generator's `NavBar` unconditionally renders `<SearchBox>`, even when the site was generated without the `orama-db` generator. In that case there's no `/orama-db.json` index for `useOrama` to fetch, so the search bar appears in the UI but is entirely non-functional (typing into it never returns results).

## Change
- `src/generators/web/utils/config.mjs`: the `#theme/config` virtual module now also exports `hasSearch`, computed from the full run configuration's `target` list (`getConfig().target.includes('orama-db')`).
- `src/generators/web/ui/components/NavBar.jsx`: imports `hasSearch` from `#theme/config` and only renders `<SearchBox>` when it's `true`.

This mirrors how other optional, generator-dependent UI is already gated, and keeps the change scoped to the two files that need it.

## Testing
- Read through `src/generators/web/utils/__tests__/config.test.mjs`: it only exercises `buildVersionEntries`, `buildPageList`, and `buildLanguageDisplayNameMap`, none of which are touched by this change, so the existing suite continues to pass as-is.
- Verified by inspection that `getConfig()` (no argument) returns the full merged configuration object (including `target`), while `getConfig('web')` returns only the per-generator config that `createConfigSource` was already destructuring - so `hasSearch` reads from the right place.

Closes #891

### DIFF
--- a/src/generators/web/ui/components/NavBar.jsx
+++ b/src/generators/web/ui/components/NavBar.jsx
@@ -6,14 +6,14 @@ import GitHubIcon from '@node-core/ui-components/Icons/Social/GitHub';
 import SearchBox from './SearchBox';
 import { useTheme } from '../hooks/useTheme.mjs';
 
-import { repository } from '#theme/config';
+import { hasSearch, repository } from '#theme/config';
 import Logo from '#theme/Logo';
 
 /**
  * NavBar component that displays the headings, search, etc.
  */
 export default ({ metadata }) => {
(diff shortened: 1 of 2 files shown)


avivkeller (changes requested):
You can't read the full config from within a generator. You can only read the config for that generator

PR #894, opened 2026-07-11
fix(web): hide search without Orama data

### LINKED ISSUE #891: `web` generator should not render the search bar if `orama-db` is not enabled

w3lld1:
## Description

Only render the web documentation search control when `orama-db` is one of the requested generator targets. This avoids showing a search UI that cannot return results when users generate only the `web` output.

The generator orchestration context now supplies the requested targets without making the web config source read the full run configuration. The web generator derives a build-time `searchEnabled` flag and supports an explicit `web.showSearchBar` override for deployments that provide search data separately. The navigation conditionally renders `SearchBox`, with regression coverage for both `web`-only and `web` + `orama-db` generation.

A patch changeset is included because this changes published web-generator behavior.

## Validation

- `node --test --experimental-test-module-mocks src/generators/web/__tests__/generate.test.mjs` (Node.js 24.18.0)
- `node --run test` — 514 tests passed (Node.js 24.18.0)
- `node --run lint` — no errors; 2 pre-existing warnings in `useOrama.mjs`
- `node --run format:check`
- `git diff --check`

## Related Issues

Fixes #891

### Check List

### DIFF
--- a/.changeset/lazy-tools-search.md
+++ b/.changeset/lazy-tools-search.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': patch
+---
+
+Hide the web documentation search control unless the Orama database generator is enabled.
--- a/src/generators.mjs
+++ b/src/generators.mjs
@@ -73,7 +73,9 @@ const createGenerator = () => {
           ? createParallelWorker(generatorName, pool, configuration)
           : Promise.resolve(null);
 
-        const result = await generate(dependencyInput, await worker);
+        const result = await generate(dependencyInput, await worker, {
+          target: configuration.target,
+        });
 
         // For streaming generators, "Completed" is logged when the cache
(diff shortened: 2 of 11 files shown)


avivkeller (dismissed):
You shouldn't read the full config from within the generator if it can be avoided

bjohansebas (commented):
think that's fine. If there's no data, why show the search? It could even help with debugging. I haven't looked through the changes yet, but in my opinion, that approach would be fine.

w3lld1:
Thanks for the review. I pushed a follow-up that avoids reading the full run config from the web config source: the orchestrator now passes requested targets through a small generator context. I also added the suggested `web.showSearchBar` override for shared/external search databases and documented it. The focused web tests, full 514-test suite, lint (no errors), formatting, and diff checks all pass on Node.js 24.18.0.

PR #900, opened 2026-07-11
Render web search only when enabled

### LINKED ISSUE #891: `web` generator should not render the search bar if `orama-db` is not enabled

Minghan2005:
## Summary

- derive the web generator's `showSearchBar` setting from whether `orama-db` is targeted
- preserve explicit `showSearchBar: true` and `showSearchBar: false` overrides
- conditionally render the navigation search UI
- document and type the new option
- cover both configuration behavior and generated HTML

## Why

The web navigation always rendered its search control, even when the generator pipeline did not build an Orama database. In that configuration the control was visible but could not return results.

The resolved configuration now enables search automatically when `orama-db` is part of the target list. Projects that share a remote database or intentionally hide search can still override the derived default explicitly.

## Validation

- `node --test --experimental-test-module-mocks "src/**/*.test.mjs"` (516 tests passed)
- `eslint . --no-warn-ignored` (0 errors; 2 existing warnings in `useOrama.mjs`)
- `prettier --check .`

Closes #891

### DIFF
--- a/src/generators/web/README.md
+++ b/src/generators/web/README.md
@@ -13,6 +13,7 @@ The `web` generator accepts the following configuration options:
 | `project`         | `string`  | `'Node.js'`                                   | Project name used in page titles and the version selector                                                   |
 | `title`           | `string`  | `'{project} v{version} Documentation'`        | Title template for HTML pages (supports `{project}`, `{version}`)                                           |
 | `useAbsoluteURLs` | `boolean` | `false`                                       | When `true`, all internal links use absolute URLs based on `baseURL`                                        |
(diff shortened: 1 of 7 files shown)


Minghan2005:
I hope you pass

Minghan2005:
I’ve rebased this branch onto the latest `main`, resolved the README option-table conflict while retaining both `remoteConfigUrl` and `showSearchBar`, and force-pushed the result. The full test suite passes locally (516 tests), as do formatting and lint checks (only two pre-existing warnings). The new Actions workflows are waiting for maintainer approval; could you please approve them when convenient?

PR #911, opened 2026-07-17
chore(banner): add opt-out

avivkeller:
Ref: https://github.com/nodejs/nodejs.org/pull/9008

Implements the `onClose` logic needed by https://github.com/nodejs/nodejs.org/pull/9008.


.changeset/opt-out-banners.md
```diff
@@ -0,0 +1,5 @@
+---
+'@node-core/doc-kit': patch
```
bmuenzenmeyer: this is a breaking change 

https://github.com/nodejs/learn/blob/main/components/Layout/index.jsx#L3

```suggestion
'@node-core/doc-kit': major
```

bmuenzenmeyer (changes requested):
Functionally works, but I want to align on how to move through this change management.

.changeset/opt-out-banners.md
avivkeller: My concern with making this a major is that we will have tens of doc-kit releases as we adapt the code base. we will need to make a lot of changes (which may be breaking) for Node.js Core to adopt this project.

I think we either:
1. Accept that doc-kit will have many major releases throughout it's development
2. Wait for Node Core to adopt this project, then begin following semver
3. Make all of the changes required by node core, and any other breaking changes (e.g. monorepo) and release doc-kit 2.0, begin following semver with that release

Not sure?

avivkeller: Personally, I think we treat doc-kit v1.0 as our "beta", and make doc-kit 2.0 the first release to follow semver.

bmuenzenmeyer: i am fine with that - but we need to coordinate changes then. we dont necessarily know all of our customers. we THINK we do right now, and the list is already like 5 places

avivkeller:
```suggestion
'@node-core/doc-kit': minor
```

avivkeller:
@bmuenzenmeyer are you still blocking?

bmuenzenmeyer:
Nope, we just need to coordinate

bmuenzenmeyer (approved):

PR #913, opened 2026-07-18
feat: auto-detect default config file

### LINKED ISSUE #897: Consideration: detect `doc-kit.config.mjs` automatically

112-stack:
## Description

Automatically discover `doc-kit.config.mjs` in the current working directory when `--config-file` is omitted. An explicitly supplied config path still takes precedence.

The previous loader returned an empty configuration whenever the CLI option was absent, so the conventional filename could not be used without repeating the flag. This change adds dependency-free discovery, focused coverage for discovery/absence/explicit precedence, user-facing documentation, and a minor changeset.

## Validation

- `node --test --experimental-test-module-mocks src/utils/configuration/__tests__/index.test.mjs` — 16/16 pass
- `node --run lint` — 0 errors (2 existing warnings in `useOrama.mjs`)
- `node --run format:check` — pass
- Full Windows run: 520/523 pass. The three failures are existing path-separator assumptions in `src/generators/ast/__tests__/generate.test.mjs` and `src/generators/web/utils/__tests__/copying.test.mjs`; none touch configuration loading.

## Related Issues

Fixes #897

### Check List

- [ ] I have run `node --run test` and all tests passed. (See the Windows-only baseline failures documented above.)

### DIFF
--- a/.changeset/kind-birds-detect.md
+++ b/.changeset/kind-birds-detect.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': minor
+---
+
+Automatically load `doc-kit.config.mjs` from the current working directory when `--config-file` is omitted.
--- a/README.md
+++ b/README.md
@@ -54,9 +54,11 @@ Commands:
 
 ### `generate`
 
-You must provide either `--target` (one or more generators to run) or
-`--config-file` (which supplies the targets). Running `generate` without either
-exits with an error pointing you to the help output.
+You must provide the input and target generators through CLI options or a
+configuration file. When `--config-file` is omitted, doc-kit automatically
+loads `doc-kit.config.mjs` from the current working directory if it exists. An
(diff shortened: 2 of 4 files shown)


avivkeller:
Please look at the feedback given on https://github.com/nodejs/doc-kit/pull/901 before sending an AI agent to rush in and complete the work incorrectly

avivkeller (changes requested):
Blocking, see above

112-stack:
Thanks for pointing this out,  You're right, I should have reviewed #901 and its discussion before opening this PR.

#913 duplicates that active work and only handles `doc-kit.config.mjs` inside `loadConfigFile`, while the feedback on #901 calls for resolving the full supported set (`.js`, `.cjs`, `.mjs`, `.ts`, `.cts`, and `.mts`) in a defined order before calling the loader.

I'm closing this PR rather than competing with the existing contributor's work. Sorry for the noise.

PR #919, opened 2026-07-18
fix: preserve deprecation codes in table of contents

### LINKED ISSUE #917: Show Dep Code in ToC
The table of contents for deprecations doesn't include the deprecation ID, making it harder to locate the appropriate detail section on https://nodejs-api-docs-tooling.vercel.app/deprecations.html

image

---
Compare to https://nodejs.org/api/deprecations.html

image

_Originally posted by @MikeMcC399 in https://github.com/nodejs/node/issues/62045#issuecomment-5010103983_

MFA-G:
## Description

On the deprecations page, sidebar/ToC labels are built by `headingLabel` in `src/generators/jsx-ast/utils/buildBarProps.mjs`, which strips any leading `<something>:` prefix (meant for prefixes like `Class:`). That rule also strips deprecation codes, so `DEP0001: http.OutgoingMessage.prototype.flush` renders in the ToC as just `http.OutgoingMessage.prototype.flush`, making it hard to locate a deprecation by its ID.

This PR adds a narrow exception: headings matching `DEP\d+:` keep their full text (backticks removed) as the label, so the deprecation code shows up in the ToC — matching the behavior of the legacy nodejs.org deprecations page.

## Validation

- New unit test in `src/generators/jsx-ast/utils/__tests__/buildBarProps.test.mjs`: `DEP0001: \`http.OutgoingMessage.prototype.flush\`` now yields the label `DEP0001: http.OutgoingMessage.prototype.flush`.
- `node --run test` — 523 tests pass, 0 fail.
- `node --run lint` — no errors (2 pre-existing warnings in `src/generators/web/ui/hooks/useOrama.mjs`, untouched by this PR).
- `node --run format:check` — clean.

## Related Issues

Fixes #917

### Check List

### DIFF
--- a/src/generators/jsx-ast/utils/buildBarProps.mjs
+++ b/src/generators/jsx-ast/utils/buildBarProps.mjs
@@ -9,9 +9,6 @@ import { TOC_MAX_HEADING_DEPTH } from '../constants.mjs';
 // rather than the full signature.
 const FUNCTION_HEADING_TYPES = new Set(['method', 'ctor', 'classMethod']);
 
-// Deprecation codes are part of the heading label, not a generic type prefix.
-const DEPRECATION_HEADING_REGEX = /^DEP\d+:/;
-
 /**
  * Generate a combined plain text string from all MDAST entries for estimating reading time.
  *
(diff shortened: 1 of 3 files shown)


AugustinMauroy (approved):

src/generators/jsx-ast/utils/buildBarProps.mjs
avivkeller: Can't we just modify this RegEx to exclude `DEPXXXX`?

MFA-G: Updated as suggested: the generic prefix-removal regex now uses a negative lookahead for `DEP\d+:`, so the deprecation case stays within the existing formatting pipeline instead of using a separate branch. The existing regression test still passes, along with the full test suite, lint, and formatting checks.

avivkeller (approved):

PR #921, opened 2026-07-19
fix(oxc): fall back to wasm

avivkeller:
As titled

### DIFF
--- a/.changeset/oxc-fallback.md
+++ b/.changeset/oxc-fallback.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': patch
+---
+
+Fall back to WASM when a native OXC binding is unavailable
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@node-core/doc-kit",
-  "version": "1.4.1",
+  "version": "1.4.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@node-core/doc-kit",
-      "version": "1.4.1",
+      "version": "1.4.3",
       "dependencies": {
         "@actions/core": "^3.0.0",
         "@heroicons/react": "^2.2.0",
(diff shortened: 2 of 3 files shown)


avivkeller:
cc @boshen howdy from Node.js! Any idea on how to allow a WASM fallback without npm throwing invalid architecture errors?

avivkeller:
In the meantime, I'm going to swap us for a WASM parser like swc or yuku

AugustinMauroy (approved):

AugustinMauroy:
https://stackblitz.com/edit/oxc-parser?file=package.json
https://npmx.dev/package-code/oxc-parser/v/0.140.0/src-js%2Fwasm.js

miss click and I cannot discard my approval

Boshen:
> cc @Boshen howdy from Node.js! Any idea on how to allow a WASM fallback without npm throwing invalid architecture errors?

`@oxc-parser/binding-wasm32-wasi` sets `cpu` to `wasm32` so adding it as a dependency fails. Should it be `optionalDependency`?

avivkeller:
That wouldn't help, would it? Our consumer (node core) runs `npm ci`, so it wouldn't install it if it were an optionalDependency

Boshen:
Wait ... why do you need this fallback when `oxc-parser` provides `@oxc-parser/binding-wasm32-wasi` as an `optionalDependencies` (along with all other builds).

Which architecture are we missing, I can add those instead.

avivkeller:
> Wait ... why do you need this fallback when `oxc-parser` provides `@oxc-parser/binding-wasm32-wasi` as an `optionalDependencies` (along with all other builds).

I assumed (possibly incorrectly) that by manually installing the WASM binding, oxc would 'fall back' to it if no bindings were found.

> Which architecture are we missing, I can add those instead.

Thank you, that would be great (if it's not too much work, of course)!

In our CI, the following runners failed:
- `smartos23-x64` (I'm guessing this is `smartos` and `x64` for `os` and `arch`, respectively)
- `aix72-power9` (I'm guessing this is `aix` and `ppc64` for `os` and `arch`, respectively)
- `aix73-power9` (I'm guessing this is `aix` and `ppc64` for `os` and `arch`, respectively)

If this is too difficult, we can also throw an error that doc generation is not available on these machines (cc @nodejs/web-infra wdyt?). Although, adding these architectures would allow us to use

AugustinMauroy:
@Boshen can we simply import was build of oxc as `oxc-parser/wasm` ?

avivkeller:
I don't think so, `@oxc-parser/binding-wasm32-wasi` only installs on specific machines

AugustinMauroy:
That seems something that we can fix in upstream

PR #922, opened 2026-07-19
fix: oxc-parser to yuku-parser

avivkeller:
Swaps our usage of OXC with a WASM alternative


AugustinMauroy:
not fan of yuk-parser too recent and zig

araujogui:
What's the advantage of yuku-parser compared to oxc?

avivkeller:
It doesn't crash the Node.js CI by requiring a native binding for niche platforms :-)

bmuenzenmeyer:
I feel the same as above - never heard of it and want to more thoroughly vet something going into core CI

avivkeller:
I can switch to swc, but Yuku claims to be much faster:
image

avivkeller:
IMO there's nothing wrong with using a smaller package, [it's spec compliant](https://github.com/yuku-toolchain/parser-test-suite), and this gives us the opportunity to work with other maintainers :-)

bjohansebas:
My only concern is who's behind the project—I mean, whether it's going to be maintained in the long run. At least with OXC, it's pretty clear it'll continue to be maintained, and the same is true for SWC because companies have a vested interest in it.

Whether it's written in another language doesn't really matter to me. What matters is who's maintaining it and what their long-term plans are.

bmuenzenmeyer:
I tried to check out their openssf scorecard, only to find https://github.com/yuku-toolchain/yuku/security - agree it looks fast, but we need HIGH SCRUTINY on every dep, in my opinion

avivkeller:
So do we want swc?

bmuenzenmeyer:
sorry if you felt pressured to close this. i sometimes crave a little more context and slowness. i saw some references to why this was needed, but it is still foggy to me if it can be solved within oxc or not

avivkeller:
No, no, It's fine. We just need a parser that's WASM rather than native, since not all machines have native bindings

avivkeller:
I'll open a followup to replace oxc with swc

PR #926, opened 2026-07-20
remove: better usage of oxc ecosystem

AugustinMauroy:
## Description

**main**: +-200ms for api-link test 
**with changes**: 150ms api-link for test 

So there aren't perf gap we just win on the amount of dependency 

also since `oxc-parser` didn't have `loc` I had to write small code that do same job once oxc-parser will have loc we will be able to come back

also use `esrap` recommend by oxc https://oxc.rs/docs/guide/usage/parser.html#print which is lighter 

## Validation

test should pass

## Related Issues

NA

### Check List


overlookmotel:
Hi. I'm one of Oxc's core team.

I'd recommend using `Visitor` provided by `oxc-parser` over `oxc-walker` if you can. `oxc-walker` has some features that `Visitor` doesn't, but `Visitor` is much faster, so it's preferred if you can use it.

https://npmx.dev/package/oxc-parser#user-content-example

AugustinMauroy:
thanks @overlookmotel for your input there. and any more info around `loc` ?

src/generators/api-links/utils/findDefinitions.mjs
```diff
@@ -34,112 +41,112 @@ function handleAssignmentExpression(node, nameToLineNumberMap, exports) {
...
 function handleFunctionDeclaration(
   node,
   basename,
   nameToLineNumberMap,
-  exports
+  exports,
+  sourceText
 ) {
   if (!exports.identifiers.includes(node.id.name)) {
-    // Function isn't exported, not relevant to us
```
avivkeller, line 101: Why are all these comments removed?

src/generators/jsx-ast/generate.mjs
```diff
@@ -1,4 +1,5 @@
-import { jsx, toJs } from 'estree-util-to-js';
+import { print } from 'esrap';
+import tsx from 'esrap/languages/tsx';
```
avivkeller, line 2: Does estree-util-to-js not support oxc?

src/generators/jsx-ast/generate.mjs
```diff
@@ -46,7 +47,13 @@ export async function processChunk(slicedInput, itemIndices) {
 
     const content = await buildContent(entries, head);
 
-    const { value: code } = toJs(content, { handlers: jsx });
+    const { code } = print(
+      content,
+      tsx({
+        // SUPER IMPORTANT: use double instead roldown will fail on certain unicode
```
avivkeller, line 53: Huh?

src/generators/jsx-ast/generate.mjs
avivkeller: If we change that, it should be a separate discussion, IMO

src/generators/api-links/utils/findDefinitions.mjs
AugustinMauroy: like mister bean said "magics" I mean it's an error from my side while working on that

src/generators/jsx-ast/generate.mjs
AugustinMauroy: yeah is it but esrap is much more smaller 183,9 ko and `estree-util-to-js` is 529,6 ko. and it's look little bit quicker (based on tests I didn't get why benchmark GH comment is updated)

src/generators/jsx-ast/generate.mjs
AugustinMauroy: yeah if you didn't set this option it's will break. for certain unicode

```
[PARSE_ERROR] Invalid Unicode escape sequence
 -  in \0virtual:crypto.jsx at 44577..44578
 ```

avivkeller (changes requested):
We can't rely on oxc until they provide native bindings for all the platforms in Node.js's CI, or we don't build the docs on those platforms.

PR #930, opened 2026-07-21
perf: fix render-blocking CSS and font display

moshams272:
## Description

As shown in the attached Lighthouse performance reports, the generated HTML templates have Render-Blocking issues caused by the main `styles.css` and the Google Fonts CSS files. Furthermore, the Google Fonts lacked display=swap, causing a Flash of Invisible Text (FOIT).

image

image

This PR improves performance (both web and legacy) by:

- Implementing a high-performance `rel="preload"` pattern: that instructs the browser to download the file with low priority. The `onload` script flips the attribute back to stylesheet once downloading is complete, applying the styles smoothly.

- Using `display=swap` on Google Fonts to ensure text is immediately visible with the default text till the font is downloaded, then apply that.

## Validation

Tested on `webpack/webpack-doc-kit`.

**Before:**
image

**After:**
image

## Related Issues

None

### Check List

### DIFF
--- a/src/generators/legacy-html/template.html
+++ b/src/generators/legacy-html/template.html
@@ -7,10 +7,32 @@
     <meta name="nodejs.org:node-version" content="__VERSION__" />
     <title>__SECTION__ | Node.js __VERSION__ Documentation</title>
     <link
-      rel="stylesheet"
-      href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic&display=fallback"
+      rel="preload"
+      href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic&display=swap"
+      as="style"
+      onload="
+        this.onload = null;
+        this.rel = 'stylesheet';
+      "
     />
-    <link rel="stylesheet" href="assets/style.css" />
+    <noscript>
+      <link
+        rel="stylesheet"
+        href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic&display=swap"
+      />
+    </noscript>
+    <link
+      rel="preload"
+      href="assets/style.css"
+      as="style"
+      onload="
+        this.onload = null;
+        this.rel = 'stylesheet';
+      "
+    />
+    <noscript>
+      <link rel="stylesheet" href="assets/style.css" />
+    </noscript>
     <link rel="canonical" href="https://nodejs.org/api/__FILENAME__.html" />
     <script async defer src="assets/api.js" type="text/javascript"></script>
     <script>
--- a/src/generators/web/template.html
+++ b/src/generators/web/template.html
@@ -5,7 +5,10 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <title>${title}</title>
-  <link rel="stylesheet" href="${root}styles.css" />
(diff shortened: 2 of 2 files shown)


src/generators/web/template.html
```diff
@@ -5,16 +5,21 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <title>${title}</title>
-  <link rel="stylesheet" href="${root}styles.css" />
+  <link rel="preload" href="${root}styles.css" as="style" onload="this.onload=null;this.rel='stylesheet'">
+  <noscript>
+    <link rel="stylesheet" href="${root}styles.css">
+  </noscript>
```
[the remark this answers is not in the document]
moshams272: We can use library `critters`, WDUT?!

src/generators/legacy-html/template.html
```diff
@@ -7,10 +7,32 @@
...
+      href="assets/style.css"
+      as="style"
+      onload="
+        this.onload = null;
+        this.rel = 'stylesheet';
+      "
+    />
+    <noscript>
+      <link rel="stylesheet" href="assets/style.css" />
+    </noscript>
```
avivkeller, line 35: IMO we shouldn't mix CSS and JavaScript in this way. CSS shouldn't rely on the JavaScript capabilities of a given machine.

Unfortunately, I think we need to just eat the perf issue. We might be able to reduce the size of CSS in doc-kit, perhaps, but I won't make you work on that, it's our fault, and not something that's particularly 'fun' or easy to fix.

moshams272: We can use library `critters`, It extracts the Critical CSS and inlines it directly into a `<style>` tag in the generated HTML. This means the initial render relies on zero client-side JS and has no render-blocking network requests, WDUT?!

moshams272: I searched on it since few hours, https://github.com/nodejs/doc-kit/pull/930#discussion_r3624999471

avivkeller:
I think a better alternative to this is for us to look into improving our CSS at it's core, not ways around it's current state. Thank you anyway :-). We (doc-kit maintainers) can look into this.

src/generators/legacy-html/template.html
```diff
@@ -7,10 +7,32 @@
...
     <link
-      rel="stylesheet"
-      href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic&display=fallback"
+      rel="preload"
+      href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic&display=swap"
+      as="style"
+      onload="
+        this.onload = null;
+        this.rel = 'stylesheet';
+      "
```
bjohansebas, line 16: we should preload the Google Fonts stylesheet obviously without any JavaScript, just using the standard preload/prefetch approach. I'm still surprised we're using Google Fonts instead of hosting them ourselves, but that's a separate discussion.

moshams272:
> I think a better alternative to this is for us to look into improving our CSS at it's core, not ways around it's current state. Thank you anyway :-). We (doc-kit maintainers) can look into this.

NP, you're right, BTW.

src/generators/legacy-html/template.html
moshams272: Self-hosting the fonts would be the ideal solution 👀

PR #931, opened 2026-07-22
fix(llms-txt): skip empty headings

avivkeller:
Ref: https://github.com/webpack/webpack-doc-kit/pull/208

### DIFF
--- a/src/generators/llms-txt/generate.mjs
+++ b/src/generators/llms-txt/generate.mjs
@@ -18,7 +18,7 @@ export async function generate(input) {
   const template = await readFile(config.templatePath, 'utf-8');
 
   const apiDocsLinks = input
-    .filter(entry => entry.heading.depth === 1)
+    .filter(entry => entry.heading.depth === 1 && entry.heading.data.text)
     .map(entry => `- ${buildApiDocLink(entry, config)}`)
     .join('\n');
 


bmuenzenmeyer (changes requested):
Is this a violation of the spec? https://nodejs-doc-kit.vercel.app/specification.html#421-depth-1

avivkeller:
@bmuenzenmeyer While consumers shouldn't do this as it's violation of the spec, it would be fine for us to support this non-spec syntax as a fallback?

bmuenzenmeyer:
Fair, but reminder that we now must be more thoughtful as we have public-facing docs. Can you update them as part of this work? 

https://nodejs-doc-kit.vercel.app/getting-started.html

image

bmuenzenmeyer (approved):
you have my approval after we update the spec or docs, or both
i trust you to land it then

avivkeller:
> Fair, but reminder that we now must be more thoughtful as we have public-facing docs. Can you update them as part of this work?
> 
> [nodejs-doc-kit.vercel.app/getting-started.html](https://nodejs-doc-kit.vercel.app/getting-started.html)
> 
> image

This statement is still correct. This change is making the LLM generator _not_ omit when a heading is not found. The spec requires a heading, so when a heading is not present, we "skip". I don't think this requires a change to either?

PR #932, opened 2026-07-22
chore: stop versioning generators

avivkeller:
consumers should refer to the main doc-kit version


bmuenzenmeyer (approved):
changeset?

avivkeller:
> changeset?

Is there a need for one, we don't actually rely on these values anywhere, so the users wouldn't even notice this change

AugustinMauroy (approved):

PR #934, opened 2026-07-22
fix(ast): swcify

avivkeller:
OXC does not work on all node.js CI runners. Until they support all our runners, rely on a WASM-based alternative, `@swc/wasm`


AugustinMauroy:
why wasm build work for SWC but not for OXC ? it's should

avivkeller:
> why wasm build work for SWC but not for OXC ? it's should 

OXC's WASM only installs on specific machines, whereas swc's installs on _all_.

bmuenzenmeyer (approved):

PR #937, opened 2026-07-23
chore(build): implement bundler adapter contract

avivkeller:
**Note**: I used a bit of AI assistance with this PR to resolve some issues with Rolldown->Vite (e.g. virtual modules)

By using Vite (which uses Rolldown internally) in the `web` generator, we no longer need to maintain a decent amount of custom overhead, including:
- Custom resolvers
- Custom CSS transformer (now that we don't use one, we can import CSS from external packages.[^1])

Additionally, by utilizing Vite's handling of HTML sources, we can reduce our memory by using `write: true`, and outputting the HTML as they generated (and removing them from memory). This doesn't have _a lot_ of memory, but it does save a look 7-10%.

[^1]: There was a reason we didn't do it earlier, IIRC it wasn't working, but I'm not 100% sure. Regardless, it works now.


avivkeller:
> -635.66 MB (-10.90%)

That's a good sign

avivkeller:
Screenshot 2026-07-23 at 2 38 58 PM

Perfomance is up to 91% :-)

avivkeller:
Bump @nodejs/web

bjohansebas (approved):
LGTM. I'm just not entirely sure if it works with webpack, since the whole idea was for it to be agnostic.

avivkeller:
> LGTM. I'm just not entirely sure if it works with webpack, since the whole idea was for it to be agnostic.

It should. The bundler contract should work with any bundler that can read our inputs and return our outputs. Vite is the default, though.

AugustinMauroy (approved):

src/generators/web/bundlers/vite.mjs
```diff
@@ -0,0 +1,398 @@
...
+  generateBundle: {
+    order: 'post',
+    /**
+     * @param {object} _
+     * @param {Record<string, object>} bundle
+     */
+    async handler(_, bundle) {
+      await Promise.all(
+        Object.values(bundle)
+          .filter(
```
ovflowd, line 102: nit: can both the filter() and map() callbacks become dedicated fns?

Can even be smth like

```tsx
async handler(_, bundle) {
  const filterAssetsFn = ...

  const mapAssetsFn = ...

  // rest of code
```

src/generators/web/bundlers/vite.mjs
```diff
@@ -0,0 +1,398 @@
...
+ * @param {'client'|'server'} environment
+ */
+const createBuildEntries = (codeMap, environment) => {
+  const input = {};
+  const sources = new Map();
+
+  for (const [fileName, code] of codeMap) {
+    const name = basename(fileName, '.jsx');
+    const id =
+      environment === 'client'
```
ovflowd, line 133: nit: inline comment quickly explaining this

src/generators/web/bundlers/vite.mjs
```diff
@@ -0,0 +1,398 @@
...
+      ? String(webConfig.baseURL).replace(/\/?$/, '/')
+      : './',
+    appType: 'custom',
+    publicDir: false,
+    clearScreen: vite.clearScreen ?? false,
+    logLevel: vite.logLevel ?? 'warn',
+
+    // Virtual entries must resolve before user plugins, while user plugins can
+    // still transform every module and generated HTML page.
+    plugins: [
```
ovflowd, line 184: Could we use .concat()/.push() instead of spreads? (and said concats could happen before the `return {` statement, just to keep it clear

src/generators/web/bundlers/vite.mjs
```diff
@@ -0,0 +1,398 @@
...
+            ? {
+                entryFileNames: '[name].mjs',
+                chunkFileNames: 'assets/[name]-[hash].mjs',
+              }
+            : {}),
+        },
+      },
+    },
+
+    ...(server
```
ovflowd, line 260: imo: we should append this to the config file if server env;

So having a :

```tsx
const baseConfig = ...

if (server) {

}

return baseConfig;
```

src/generators/web/bundlers/vite.mjs
```diff
@@ -0,0 +1,398 @@
...
+  config,
+  vite = {},
+}) => {
+  const { input, sources } = createBuildEntries(entries, 'server');
+
+  for (const [id, code] of Object.entries(virtualImports)) {
+    sources.set(id, code);
+  }
+
+  const temporaryDirectory = await createTemporaryDirectory();
```
ovflowd, line 307: Can you document what these temp dirs are for?

ovflowd: And is this what alleviates a bit of memory? By offloading to disk?

ovflowd (approved):
SGTM, but could you address my comments?

Also can you verify vite will work in node-core? For all our runners?

src/generators/web/bundlers/vite.mjs
avivkeller: Why? What's wrong with spreads

src/generators/web/bundlers/vite.mjs
avivkeller: https://github.com/nodejs/doc-kit/pull/937#discussion_r3667704183

src/generators/web/bundlers/vite.mjs
avivkeller: What's wrong with inlining as is? These aren't particularly complex

avivkeller:
> Also can you verify vite will work in node-core? For all our runners?

It'll work in all the runners that currently support building the `web` generator (i.e. runners without support for specific native bindings will fail, but they wouldn't have enough memory to run the web generator anyway. IMO it's worth the tradeoff of more nonsupport for specific niche runners for faster and lower memory development for all the rest)

src/generators/web/bundlers/vite.mjs
avivkeller: Yes, this saves _some_ memory, but the majority of the memory saves is from Vite writing the JS entrypoints directly to the outdir (only the HTML is in memory, not the HTML+JS)

PR #939, opened 2026-07-25
feat(config): cosmiconfig

### LINKED ISSUE #897: Consideration: detect `doc-kit.config.mjs` automatically

avivkeller:
Fixes https://github.com/nodejs/doc-kit/issues/897 by adding `cosmiconfig` as a dependency.


bmuenzenmeyer (approved):
i didnt run it locally, but LGTM

bjohansebas (approved):

PR #944, opened 2026-07-26
fix: do not render headings with no content

moshams272:
## Description

Currently, if an MDX page lacks an ATX H1 (e.g., when the `h1` is built into a custom React component), `doc-kit` automatically injects an empty `h1` to satisfy AST structural requirements and metadata generation.
 
However, this leads to an issue for MDX: if it renders its own custom `h1` in a React component, the final HTML ends up with **two `h1` tags** (one from `doc-kit`, one from the component), breaking SEO and Accessibility.

 The current workaround involves forcing a `# Heading` in markdown and visually hiding it via CSS (`sr-only`), which is a hack.
 
This PR still injects an empty `h1`, but it doesn't render the empty headings in general to DOM.

## Validation

Tested on `webpack-doc-kit`:

**Before:**
image

**After:**
image

## Related Issues

None

### Check List

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -117,7 +117,7 @@ reproduce literal output.
 A conforming [document][term-document] MUST contain elements in the following
 order. Items marked OPTIONAL MAY be omitted entirely.
 
-1. **Title heading** (OPTIONAL) - An [ATX heading][cm-atx-heading] at depth 1.
+1. **Title heading** (OPTIONAL, RECOMMENDED) - An [ATX heading][cm-atx-heading] at depth 1.
 2. **[Simple directives][§6.2]** (OPTIONAL) - Zero or more
    [HTML comment][cm-html-comment] directives providing
    [document][term-document]-level metadata. These MUST immediately follow
@@ -173,7 +173,7 @@ NOT be used.
 
 #### 4.2.1. Depth 1
 
-The [document][term-document] title. Exactly one depth-1 heading MUST appear
+The [document][term-document] title. Exactly one depth-1 heading SHOULD appear
 per document.
 
 #### 4.2.2. Depth 2
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -117,8 +117,7 @@ reproduce literal output.
 A conforming [document][term-document] MUST contain elements in the following
 order. Items marked OPTIONAL MAY be omitted entirely.
 
-1. **Title heading** (REQUIRED) - Exactly one [ATX heading][cm-atx-heading]
-   at depth 1.
+1. **Title heading** (OPTIONAL) - An [ATX heading][cm-atx-heading] at depth 1.
 2. **[Simple directives][§6.2]** (OPTIONAL) - Zero or more
    [HTML comment][cm-html-comment] directives providing
    [document][term-document]-level metadata. These MUST immediately follow
--- a/packages/core/src/generators/metadata/utils/parse.mjs
+++ b/packages/core/src/generators/metadata/utils/parse.mjs
@@ -87,7 +87,7 @@ export const parseApiDoc = ({ path, tree, mdx = false }, typeMap) => {
   // Handles the normalisation URLs that reference to API doc files with .md extension
   visit(tree, UNIST.isMarkdownUrl, node => visitMarkdownLink(node));
 
-  // If the document has no headings but it has content, we add a fake heading to the top
+  // If the document has no headings but it has content, we add a fake heading to the top, it will not be rendered
   if (headingNodes.length === 0 && tree.children.length > 0) {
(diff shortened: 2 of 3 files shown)


avivkeller:
Can you:
1. Make this not an option, we shouldn't render elements with no content in any case
2. Patch the docs + spec to remove the note about titleless pages?

moshams272:
> Can you:
> 
> 1. Make this not an option, we shouldn't render elements with no content in any case
> 2. Patch the docs + spec to remove the note about titleless pages?

you mean for headings only, right?

avivkeller:
yes sorry

src/generators/metadata/utils/parse.mjs
```diff
@@ -87,7 +87,7 @@ export const parseApiDoc = ({ path, tree, mdx = false }, typeMap) => {
   // Handles the normalisation URLs that reference to API doc files with .md extension
   visit(tree, UNIST.isMarkdownUrl, node => visitMarkdownLink(node));
 
-  // If the document has no headings but it has content, we add a fake heading to the top
+  // If the document has no headings but it has content, we add a fake heading to the top, it will not be rendered
   if (headingNodes.length === 0 && tree.children.length > 0) {
     tree.children.unshift(createTree('heading', { depth: 1 }, []));
   }
```
avivkeller, line 93: Should we just remove this statement... Why are we adding a fake heading in the first place?

Then, `createHeadings` won't need a fast path

bmuenzenmeyer:
this feels very related to notions within https://github.com/nodejs/doc-kit/pull/931

avivkeller:
> this feels very related to notions within https://github.com/nodejs/doc-kit/pull/931

Sort of, yes. We have one page without a heading, since we add the heading via JSX

src/generators/metadata/utils/parse.mjs
moshams272: Actually, I was thinking of that and searching for what the effect will be. Let me revise what I found.

moshams272: 1. While traversing AST, in this [file](https://github.com/nodejs/doc-kit/blob/e4283811ac3adf16cc5aaeed074738ef1c8290a2/src/generators/metadata/utils/parse.mjs#L100), `parseApiDoc` extracts content by visiting heading nodes `visit(tree, UNIST.isHeading, (headingNode, index) => {...}}`. It uses each heading as an anchor and slices the AST content between it and the next heading. If a document has no headings, the `visit` never fires, and `parseApiDoc` returns an empty array, which drops all the page content.

2. in [type.d.ts](https://github.com/nodejs/doc-kit/blob/e4283811ac3adf16cc5aaeed074738ef1c8290a2/src/generators/metadata/types.d.ts#L139), The MetadataEntry interface strictly requires heading: HeadingNode (it is not optional).

avivkeller: I see, so headings are required, I guess. In that case, we should make the default heading something like `{project} Documentation`, and then in webpack use CSS/postprocessing to remove it

moshams272: I just think of **accessibility, SEO and other projects** that will use `doc-kit`, instead of pushing elements to the DOM that serve zero visual purpose, we can handle it here.

In the end, that's your decision, guys 😅

avivkeller: Your right

docs/specification.md
```diff
@@ -221,8 +221,7 @@ reproduce literal output.
 A conforming [document][term-document] MUST contain elements in the following
 order. Items marked OPTIONAL MAY be omitted entirely.
 
-1. **Title heading** (REQUIRED) - Exactly one [ATX heading][cm-atx-heading]
-   at depth 1.
+1. **Title heading** (OPTIONAL) - An [ATX heading][cm-atx-heading] at depth 1.
```
avivkeller:
```suggestion
1. **Title heading** (OPTIONAL, RECOMMENDED) - An [ATX heading][cm-atx-heading] at depth 1.
```

avivkeller (approved):

avivkeller (approved):

moshams272:
Could u tell me what the Vercel problem is 😅

ovflowd:
> Could u tell me what the Vercel problem is 😅

No idea, seems just authorization issue. Probably because needs rebase. @moshams272 could you rebase please?

AugustinMauroy (approved):

moshams272:
> No idea, seems just authorization issue. Probably because needs rebase. @moshams272 could you rebase please?

Github things, I thought the deploy failed :)

bjohansebas (approved):

PR #947, opened 2026-07-27
fix(metadata): reset `lastIndex` of unix manual page regex before testing

btea:
`QUERIES.unixManualPage` is a global regex (`/g`), so its `lastIndex` persists between calls. `UNIST.isTextWithUnixManual` uses `.test()` on this shared instance while visiting text nodes: after a successful match, the next `.test()` starts from the end of the previous match, intermittently returning `false` for text that does contain a Unix manual page reference.

This resets `lastIndex` to 0 before each test so every node is matched from the start.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#:~:text=They%20store%20a%20lastIndex%20from%20the%20previous%20match.%20Using%20this%20internally%2C%20test()%20can%20be%20used%20to%20iterate%20over%20multiple%20matches%20in%20a%20string%20of%20text%20(with%20capture%20groups).

### DIFF
--- a/src/utils/queries/index.mjs
+++ b/src/utils/queries/index.mjs
@@ -41,8 +41,10 @@ export const UNIST = {
    * @param {import('@types/mdast').Text} text
    * @returns {boolean}
    */
-  isTextWithUnixManual: ({ type, value }) =>
-    type === 'text' && QUERIES.unixManualPage.test(value),
+  isTextWithUnixManual: ({ type, value }) => {
+    QUERIES.unixManualPage.lastIndex = 0; // Reset the lastIndex to ensure proper matching
+    return type === 'text' && QUERIES.unixManualPage.test(value);
+  },
 
   /**
    * @param {import('@types/mdast').Link} link
--- a/.changeset/unix-manual-regex-lastindex.md
+++ b/.changeset/unix-manual-regex-lastindex.md
@@ -1,0 +1,7 @@
+---
+'@node-core/doc-kit': patch
+---
+
+Fix `isTextWithUnixManual` intermittently missing valid Unix manual page
+references: the shared `unixManualPage` regex is global, so its `lastIndex`
+persisted between `.test()` calls during tree traversal.


src/utils/queries/index.mjs
```diff
@@ -41,8 +41,10 @@ export const UNIST = {
    * @param {import('@types/mdast').Text} text
    * @returns {boolean}
    */
-  isTextWithUnixManual: ({ type, value }) =>
-    type === 'text' && QUERIES.unixManualPage.test(value),
+  isTextWithUnixManual: ({ type, value }) => {
+    QUERIES.unixManualPage.lastIndex = 0; // Reset the lastIndex to ensure proper matching
+    return type === 'text' && QUERIES.unixManualPage.test(value);
+  },
```
avivkeller, line 47: Can you give an example of something it would miss?

avivkeller:
> @btea is attempting to deploy a commit to the **OpenJS Foundation** Team on [Vercel](https://vercel.com).
> 
> A member of the Team first needs to [authorize it](https://vercel.com/git/authorize?team=OpenJS%20Foundation&slug=openjs&teamId=team_XimxEuPDAZ4tgEUd3NwzudNC&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%2296d492ec2e21208b89911ce24552552856dcb6bc%22%7D%2C%22id%22%3A%22QmVZr23dgg68g9pgHg52PHR5Vfq3bsbTk3tb1h1fKKzPbZ%22%2C%22org%22%3A%22nodejs%22%2C%22prId%22%3A947%2C%22repo%22%3A%22doc-kit%22%7D).

@nodejs/web-admins this shouldn't happen, we only want builds on `api-docs-tooling`

MattIPv4:
> > @btea is attempting to deploy a commit to the **OpenJS Foundation** Team on [Vercel](https://vercel.com).
> > A member of the Team first needs to [authorize it](https://vercel.com/git/authorize?team=OpenJS%20Foundation&slug=openjs&teamId=team_XimxEuPDAZ4tgEUd3NwzudNC&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%2296d492ec2e21208b89911ce24552552856dcb6bc%22%7D%2C%22id%22%3A%22QmVZr23dgg68g9pgHg52PHR5Vfq3bsbTk3tb1h1fKKzPbZ%22%2C%22org%22%3A%22nodejs%22%2C%22prId%22%3A947%2C%22repo%22%3A%22doc-kit%22%7D).
> 
> @nodejs/web-admins this shouldn't happen, we only want builds on `api-docs-tooling`

The ignore step was already set to production only, I've now disabled the "Pull Request Comments" Git functionality, will see if that helps.

src/utils/queries/index.mjs
btea: I dug into this, and honestly: with the current call chain, nothing is observably missed today. `isTextWithUnixManual is` only used as a visit() predicate ([src/generators/metadata/utils/parse.mjs:142](https://github.com/nodejs/doc-kit/blob/cef45caf14cba83c77c373f59cee7f33f9f3df9b/src/generators/metadata/utils/parse.mjs#L142)), and every successful .test() is immediately followed by the visitor's node.value.replace(QUERIES.unixManualPage, ...) ([src/generators/metadata/utils/visitors.mjs:33](https://github.com/nodejs/doc-kit/blob/cef45caf14cba83c77c373f59cee7f33f9f3df9b/src/generators/metadata/utils/visitors.mjs#L33), same regex instance)

[`RegExp.prototype[@@replace]`](https://tc39.es/ecma262/#sec-regexp.prototype-@@replace), step 9, on a global regex resets lastIndex to 0. A failing .test() also resets it, so lastIndex is always 0 at the start of each test. Verified empirically: pre-fix vs post-fix predicates over all of node/doc/api/*.md produce identical matches.                                                         
                                                                               
So this is hardening against shared mutable regex state, not a fix for a currently visible bug. Happy to add a small unit test calling the predicate twice to lock it in, or close if you'd rather not carry the extra line.

avivkeller: Without a current bug this fix, I see no reason to land

PR #951, opened 2026-07-27
feat: partial hydration

### LINKED ISSUE #353: Don't pass pre-rendered components to the client
(Follow-up #285)

Currently, the web generator statically renders a large portion of the documentation, and then rehydrates it on the client. Most of this code doesn't change, and doesn't need to be rehydrated. We should find a way to avoid passing this to the client at all.

Image[^1]

[^1]: Visual from https://www.gatsbyjs.com/docs/conceptual/partial-hydration/

### LINKED ISSUE #815: Reduce memory footprint

### LINKED ISSUE #909: Possible bug on Windows 25H2
> May I ask for some browser + extension information, we've received a similar report before, but, like this one, I can't reproduce, unfortunately:
> 
> Screenshot 2026-07-15 at 10 51 34 PM

I'm seeing the issue only on Windows 11 25H2, not on Ubuntu 24.04.4 LTS:
- Google Chrome Version 150.0.7871.125 (Official Build) (64-bit) - also in Incognito mode
- Mozilla Firefox 152.0.6 (64-bit) - also in private window

_Originally posted by @MikeMcC399 in https://github.com/nodejs/node/issues/62045#issuecomment-4989568601_

avivkeller:
Uses islands from 11ty in order to give us partial hydration

### DIFF
--- a/packages/core/src/generators/web/ui/islands/runtime.mjs
+++ b/packages/core/src/generators/web/ui/islands/runtime.mjs
@@ -1,0 +1,63 @@
+import { Island } from '@11ty/is-land';
+import { h, hydrate } from 'preact';
+
+import loaders from './loaders.mjs';
+
+/**
+ * Adds the generator's component map to the registry.
+ *
+ * @param {Record<string, () => Promise<{ default: import('preact').ComponentType }>>} components
+ */
+export const registerIslands = components => Object.assign(loaders, components);
+
+/**
+ * Re-renders an island's server-rendered children as the markup they already
+ * are. Preact keeps the existing DOM because the HTML is identical, so nothing
+ * static ever has to be shipped as JavaScript.
+ *
+ * @param {{ html: string }} props
+ */
+const Slot = ({ html }) =>
+  h('island-slot', { dangerouslySetInnerHTML: { __html: html } });
+
+// Registered before any island can reach `beforeReady`: importing is-land above
+// upgrades the elements already in the document, but `Island#hydrate` awaits its
+// loading conditions first, and that await cannot resolve until this module body
+// has run to completion.
+Island.addInitType('preact', async island => {
+  const name = island.getAttribute('data-island-name');
+  const loader = loaders[name];
+
+  if (!loader) {
+    console.error(`[is-land] no component registered for "${name}"`);
+    return;
+  }
+
+  const script = [...island.children].find(child =>
+    child.matches('script[data-island-props]')
+  );
+
+  const props = script ? JSON.parse(script.textContent) : {};
+
+  // Preact hydrates by walking the container's children in order, so the props
+  // script has to go before the tree it describes is diffed against them.
+  script?.remove();
+
+  const slots = [...island.querySelectorAll('island-slot')].filter(
+    slot => slot.closest('is-land') === island
+  );
+
+  if (slots.length) {
+    props.children = slots.map(slot => h(Slot, { html: slot.innerHTML }));
+  }
+
+  try {
+    const { default: Component } = await loader();
+
+    hydrate(h(Component, props), island);
+  } catch (error) {
+    // is-land awaits this callback, so a rejection would leave the island
+    // silently stuck: never marked ready, and never reported anywhere.
+    console.error(`[is-land] "${name}" failed to hydrate`, error);
+  }
+});
--- a/packages/core/src/generators/web/utils/generate.mjs
+++ b/packages/core/src/generators/web/utils/generate.mjs
@@ -46,46 +46,30 @@ export const createImportDeclaration = (
 };
 
 /**
- * Factory function that creates server and client program generators.
- *
- * Returns two functions that wrap JSX component code:
- * - `buildClientProgram`: Wraps component for client-side hydration
- * - `buildServerProgram`: Wraps component for server-side rendering
+ * Factory function that creates the page programs.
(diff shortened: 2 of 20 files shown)


ovflowd:
@avivkeller is this fully working or on a ready state? I see the blocked label?

ovflowd:
Is it only shiki broken?

avivkeller:
> @avivkeller is this fully working or on a ready state? I see the blocked label?

It's ready for review, since the implementation on this side is done, but can't land until https://github.com/nodejs/nodejs.org/pull/9028 https://github.com/nodejs/nodejs.org/pull/9047 to

avivkeller:
Unblocked as upstream changes have landed... just dreading the rebase I'm about to make

AugustinMauroy:
idk why but when I take a page an scroll I have some lag spike in safari 26.5.2

avivkeller:
> idk why but when I take a page an scroll I have some lag spike in safari 26.5.2

I can't reproduce, probably a browser issue. If anything, this should lower spikes since it's less JS

packages/core/src/generators/web/ui/islands/runtime.mjs
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Re-renders an island's server-rendered children as the markup they already
+ * are. Preact keeps the existing DOM because the HTML is identical, so nothing
+ * static ever has to be shipped as JavaScript.
+ *
+ * @param {{ html: string }} props
+ */
+const Slot = ({ html }) =>
+  h('island-slot', { dangerouslySetInnerHTML: { __html: html } });
```
bmuenzenmeyer, line 21: is this the standard way to do this? I suppose we trust our input now, but could someone use this to render XSS into a preview?

bmuenzenmeyer (approved):
I am still learning this architecture, but don't want to be a blocker.

ovflowd:
I see some light themed items on dark mode?

image

Whereas on beta.docs.nodejs.org correctly colored?

image

avivkeller:
That's probably a bug of the ui bump

ovflowd:
> That's probably a bug of the ui bump

I believe so

packages/core/src/generators/web/ui/islands/runtime.mjs
avivkeller: > but could someone use this to render XSS into a preview?

Someone could already do that by replacing the vercel build script with whatever malicious code they want. Hence, we store nothing of value in the preview environment.

But yes, we need to `dangerouslySetInnerHTML` to insert our server dehydrated HTML

avivkeller:
Actually, it might be because in this repo we don't account for a light/dark mode in Shiki (we use `.[0]`, I'll fix it

avivkeller:
@ovflowd Fixed!

avivkeller:
Can I get approvals so I can land this :-)

packages/core/src/generators/web/ui/islands/runtime.mjs
```diff
@@ -0,0 +1,63 @@
+import { Island } from '@11ty/is-land';
```
ovflowd, line 1: Are we adding an external dependency to 11ty? I wonder we're bloating too many dependencies to doc-kit, this is included in Node.js core, so the more deps, the riskier it is.

packages/core/src/generators/web/utils/generate.mjs
```diff
@@ -110,5 +94,23 @@ export default () => {
...
+  const clientProgram = [
+    createImportDeclaration(null, resolve(ROOT, './ui/index.css')),
+
+    createImportDeclaration(
+      'registerIslands',
+      resolve(ROOT, './ui/islands/runtime.mjs'),
+      false
+    ),
+
+    `registerIslands({${componentImports
```
ovflowd, line 107: Could this map callback be a dedicated fn 🙇

packages/core/src/utils/type-annotations/hast.mjs
```diff
@@ -2,6 +2,8 @@
 
 import { highlighter } from '../highlighter.mjs';
 
+const [lightTheme, darkTheme] = highlighter.shiki.getLoadedThemes();
```
ovflowd, line 5: assumes always this order, which could change maybe?

ovflowd (approved):
SGTM, although we should be conscious about the footprint the ever-increasing dependencies are causing to doc-kit... I wish for doc-kit to be relatively small to install and not bloated.

packages/core/src/generators/web/ui/islands/runtime.mjs
avivkeller: No, `@11ty/is-land` is a dependency with no sub-deps, and is small in size.

packages/core/src/utils/type-annotations/hast.mjs
avivkeller: It shouldn't without a bump in @node-core/rehype-shiki

packages/core/src/generators/web/utils/generate.mjs
avivkeller: For such a simple function, I don't think a dedicated function is needed. We all have different preferences, I'm going to treat this is a nitpick.

PR #955, opened 2026-07-28
fix: search exiting, markdown rendering

### LINKED ISSUE #927: back tick in nav not rendering
Image

https://beta.docs.nodejs.org/module.html

https://nodejs.org/docs/latest/api/module.html

### LINKED ISSUE #943: When searching for APIs on the current module page using the quick input box, the navigation overlay does not disappear when clicking the jump button

As shown in the provided video, the mask layer does not disappear after selecting a method. Clicking on a blank area makes the mask layer disappear, but the page scrolls back to the top.

https://github.com/user-attachments/assets/bc1e8d07-1290-4a7f-bdc8-a3a784ca8f5b

avivkeller:
I know this change isn't idiomatic (hence why I wrote two changesets), I just have multiple open PRs in this repo, and I felt like it would be excessive to open two more for these simple fixes.

Fixes https://github.com/nodejs/doc-kit/issues/927 by moving the `renderLabel` utility out of `www` and into the package itself (which turns backticks into code blocks)
Fixes https://github.com/nodejs/doc-kit/issues/943 by adding an `onClick` handler to the search button, which `ESC`-es the Search (if the user isn't redirected away, that is)

### DIFF
--- a/.changeset/did-you-know-that-the-world-is-round.md
+++ b/.changeset/did-you-know-that-the-world-is-round.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': patch
+---
+
+Close Orama search when the target link is on the same page
--- a/.changeset/yes-i-did-know-that-thank-you.md
+++ b/.changeset/yes-i-did-know-that-thank-you.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': patch
+---
+
+Render markdown `code` snippets in the sidebar
--- a/packages/core/src/generators/web/ui/components/SearchBox/index.jsx
+++ b/packages/core/src/generators/web/ui/components/SearchBox/index.jsx
@@ -11,6 +11,27 @@ import styles from './index.module.css';
 import useOrama from '../../hooks/useOrama.mjs';
 import { relativeOrAbsolute } from '../../utils/relativeOrAbsolute.mjs';
 
+/**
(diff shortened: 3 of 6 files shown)


btea:
It seems the problem is only partially solved. When I search for "join" on the path page and then navigate to it, the overlay disappears, but the page content remains at the top and does not navigate to the specified method location.

avivkeller:
Bump @nodejs/web-infra

AugustinMauroy (approved):

bmuenzenmeyer:
@btea 

> It seems the problem is only partially solved. When I search for "join" on the path page and then navigate to it, the overlay disappears, but the page content remains at the top and does not navigate to the specified method location.

This appears to be working for me now using the same attempted work - can you re-verify?

bmuenzenmeyer (approved):

avivkeller:
> @btea
> 
> > It seems the problem is only partially solved. When I search for "join" on the path page and then navigate to it, the overlay disappears, but the page content remains at the top and does not navigate to the specified method location.
> 
> This appears to be working for me now using the same attempted work - can you re-verify?

It's working as of https://github.com/nodejs/doc-kit/pull/955/commits/6c3b92d3b769ff4db783a0fb0cdd462e7fc977a5, which "forces" the browser to scroll, as some won't do it automatically for whatever reason

PR #958, opened 2026-07-30
refactor(core): load generators by import specifier

avivkeller:
Generators are no longer resolved through a static registry.

Now, they're now loaded dynamically by import specifier. `--target` accepts either a built-in shorthand (`web`, `legacy-html`, …) or any import specifier that resolves to a generator module (e.g. `some-package/generator`, `./my-generator.mjs`). A generator's `dependsOn` field is likewise now a full import specifier.


.changeset/specifier-generator-loading.md
```diff
@@ -0,0 +1,11 @@
+---
+'@node-core/doc-kit': minor
```
avivkeller, line 2: This can technically be a patch, fwiw

bjohansebas (approved):

docs/generators.md
```diff
@@ -273,20 +286,23 @@ Define the generator metadata in `index.mjs`:
 
 ```javascript
 // packages/core/src/generators/streaming-generator/index.mjs
-import { createLazyGenerator } from '../../utils/generators.mjs';
```
ovflowd, line 276: How do we load them lazily or are they laoded lazily by default now?

ovflowd (approved):
Lots of changes, hard to review so rubber stamp SGTM

docs/generators.md
avivkeller: Correct, all generator imports are `await import`, so they are still lazy.

AugustinMauroy (approved):
lgmt !

PR #963, opened 2026-07-31
fix: sync URL hash when following same-page search hits

btea:
`followSearchHit` only called `scrollIntoView()` for same-page hits, leaving the address bar without a hash (or stuck on a stale one from a previously followed hit). Push the hit's hash via `history.pushState` before scrolling so the URL always reflects the jumped-to anchor.

rel https://github.com/nodejs/doc-kit/pull/955#discussion_r3686930346 cc @avivkeller 

### Check List

### DIFF
--- a/packages/core/src/generators/web/ui/components/SearchBox/index.jsx
+++ b/packages/core/src/generators/web/ui/components/SearchBox/index.jsx
@@ -21,7 +21,7 @@ const followSearchHit = event => {
     return;
   }
 
-  const target = new URL(event.currentTarget.href);
+  const { href } = event.currentTarget;
 
   event.preventDefault();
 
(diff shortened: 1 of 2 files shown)


avivkeller (approved):

packages/core/src/generators/web/ui/components/SearchBox/index.jsx
```diff
@@ -36,11 +36,15 @@ const followSearchHit = event => {
       return;
     }
 
-    requestAnimationFrame(() =>
+    requestAnimationFrame(() => {
+      // Keep the URL hash in sync with the followed hit, since
+      // `scrollIntoView` alone leaves the previous hash untouched.
+      history.pushState(null, '', target.hash || target.pathname);
+
```
avivkeller: Can we change:
```
    if (target.pathname !== window.location.pathname) {
      window.location.href = target.href;
      return;
    }
```

To `window.location.href = target.href;` and just only return if `target.pathname !== window.location.pathname`?

Something like:
```
window.location.href = target.href;
    if (target.pathname !== window.location.pathname) {
      return;
    }
```

btea: Yeah, this looks much cleaner. I've already adjusted it as suggested.

avivkeller (approved):

AugustinMauroy (approved):

PR #964, opened 2026-07-31
feat: add configuration for nav

avivkeller:
Adds `web.navigation`, which supplies the sidebar groups (`navigation.sidebar`) and the navigation bar items (`navigation.navbar`) from configuration. That way, consumers don't need to rewrite the component just to change the nav


bmuenzenmeyer:
I looked for this feature when making the getting-started section of the docs - awesome!

bmuenzenmeyer (approved):
LGTM. I was contemplating if we should alter the getting started experience to show this, but it's probably fine

AugustinMauroy (approved):

ovflowd (approved):

PR #966, opened 2026-07-31
feat(web): make the top navigation bar sticky

btea:
## Description
Keep the logo, search box, and theme toggle visible while scrolling.
- Wrap NavBar in a sticky container (z-index 40)
- Offset the sticky meta bar by --header-height so it is not covered
- Add scroll-margin-top to anchored headings below the sticky bar
- Raise Radix popups above the bar (z-index 50, overrides inline style)

It allows you to quickly search for information anywhere on the page without having to scroll to the top.
<details>
image
</details>

### Check List

### DIFF
--- a/.changeset/sticky-navigation-bar.md
+++ b/.changeset/sticky-navigation-bar.md
@@ -1,0 +1,5 @@
+---
+'@node-core/doc-kit': minor
+---
+
+Make the web top navigation bar sticky while scrolling, and offset the sticky meta bar, anchor scrolling, and Radix popups so they are not covered by it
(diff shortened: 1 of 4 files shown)


avivkeller:
Does this really need to be sticky? iirc unsticking was intentiona

btea:
> Does this really need to be sticky? iirc unsticking was intentiona

Personally, I prefer convenient and quick access to search functions in various places. In contrast, many other tool websites seem to maintain this format.

Of course, I genuinely don't know that this was a deliberate design choice. Could you explain your reasoning? Thanks!

avivkeller:
Hi! This was designed with parity for nodejs.org. Any changes regarding the sticky navbar should go through an issue in that site

btea:
It seems nodejs.org didn't add this feature before. I've already created an issue.

PR #968, opened 2026-07-31
fix: fix resolution of various type annotation edge cases

pimterry:
## Description

This fixes a cluster of related issues:

* `{` inline in normal text could break parsing, see the 3rd bullet under https://beta.docs.nodejs.org/url.html#whatwg-api
* Broken parsing of annotations like `{HTTP/2 Headers Object | HTTP/2 Raw Headers}`. The current parsing didn't handle `|` at all, it just fell back to TS, which works fine for primitives but doesn't handle these cases and printed warnings instead, leaving the links broken, e.g. `headers` option in https://beta.docs.nodejs.org/http2.html#clienthttp2sessionrequestheaders-options.
* Similar broken parsing of types like `HTTP/2 Headers Object[]` due to `[]` broke links, e.g. https://beta.docs.nodejs.org/http2.html#http2streamsentinfoheaders.
* References were highlighted as TypeScript even if they clearly weren't - this resulted in `HTTP/2 ...` showing operator highlighting for the `/` and numeric highlight for the `2` which looks weird, e.g. https://beta.docs.nodejs.org/http2.html#event-localsettings. We now parse each reference and only apply syntax highlighting on TS-parseable cases.

Some changes to other logic (detecting empty brackets and properly handling `.` separators) required en route, to avoid regressions from the expanded detection from above fixes.

## Validation

Rebuilt locally, each of the cases above now render as expected (with one caveat that `HTTP/2 Raw Headers` isn't linked yet due to https://github.com/nodejs/node/pull/64872). I've skimmed through the rest of the docs and it doesn't look like this affects any other cases.

### Check List

- [x] I have check code formatting with `node --run format:check` & `node --run lint`.


packages/core/src/generators/metadata/utils/transformers.mjs
```diff
@@ -48,6 +48,11 @@ export const lookupTypeName = (name, typeMap) => {
   return '';
 };
 
+// A dotted identifier path, e.g. `vm.Module` or `os.constants.dlopen`. Names
+// that merely contain a dot (prose, `Object.<string, string>`) must not be
+// turned into a module link.
+const MODULE_QUALIFIED_NAME = /^[A-Za-z_$][\w$]*(?:\.[A-Za-z_$][\w$]*)+$/;
```
avivkeller: Can you put this in `../constants.mjs`?

avivkeller (approved):

packages/core/src/generators/metadata/utils/transformers.mjs
avivkeller: bump

pimterry: Done :+1:

pimterry:
Fixed your comment and force-pushed to rebase & resolve the conflicts @avivkeller

packages/core/src/utils/type-annotations/syntax.mjs
```diff
@@ -3,6 +3,16 @@
...
+const SPACE = ' '.charCodeAt(0);
+
+// The type starters that are punctuation: `_` and `$` (identifiers), the
+// openers of parenthesized, tuple, object, generic and string-literal types,
+// the leading `-` of a negative numeric literal, the leading `&`/`|` of an
+// intersection or union, and `\` (a legacy escape, decoded further down the
+// pipeline). Alphanumeric and non-ASCII starters are handled by `isTypeStart`.
+const TYPE_STARTER_PUNCTUATION = new Set(
+  [...'_$([{<\'"`-&|\\'].map(character => character.charCodeAt(0))
+);
```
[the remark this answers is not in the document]
pimterry: This is just wrong I think, it's struggling to parse the string with symbols :laughing:

PR #997, opened 2026-08-04
fix: pin en-US locale in builtin MDN type links

btea:
## Description
MDN URLs without a locale prefix get redirected to the visitor's preferred language, where heading anchor IDs are translated (e.g. `#number_type` becomes `#number_类型`). The original hash fragment no longer matches any anchor after the redirect, so type links fail to jump to the target section.                                               
                                                                               
Pin the `en-US` locale on all MDN links that carry a hash fragment in the builtin type map, so the anchors resolve consistently.

### Check List


AugustinMauroy (approved):
LGMT !

avivkeller (approved):

ovflowd (approved):

btea:
@avivkeller I'd like to ask here: do we need to add this audit issue and allowScripts configuration? If so, can I submit changes, or should your team members make the changes?

image

avivkeller:
I think this is a new feature of npm. IMO we don't _need_ the scripts, and we've seen all to many recent attacks relying on those kinds of scripts, so I see no rush to add them

avivkeller:
Thanks for flagging it, however :-)

btea:
> I think this is a new feature of npm. IMO we don't _need_ the scripts, and we've seen all to many recent attacks relying on those kinds of scripts, so I see no rush to add them

Yes, this is a new npm configuration. https://docs.npmjs.com/cli/v12/commands/npm-approve-scripts
There have indeed been quite a few related attack cases recently.  Perhaps we can simply set the configuration of the relevant modules to false to avoid the above warning message appearing every time dependencies are installed.

`2 vulnerabilities (1 moderate, 1 high) 
To address all issues, run:
  npm audit fix
`

Does this part need to be processed?

PR #999, opened 2026-08-04
chore: make doc-kit friendly

avivkeller:
Ahead of RenderATL, this PR makes doc-kit user friendly with:
- `doc-kit serve`
- `doc-kit bootstrap`
- `doc-kit install`

+ updated the docs to include information about packages, these commands, and getting started.

cc @ovflowd @bmuenzenmeyer I'll open a PR when this lands to remove the Node.js-isms, figured this one was. big enough


ovflowd (approved):

packages/core/src/utils/install.mjs
ovflowd, line 1: What exactly an install command does for us?

ovflowd: I don't think we should have an install command. We should defer to package managers to install packages. Not doc-kit, I don't think any other tool does something like that.

packages/core/src/utils/bootstrap.mjs
ovflowd, line 1: Are we making some sort of CRA but for `doc-kit`?

packages/core/src/utils/install.mjs
avivkeller: We spoke in the meeting about the concern of users not being able to remember the names of packages (e.g. `@nodejs/doc-kit-generator-react`), so I figured a shorthand to install them would make sense

ovflowd (changes requested):
Actually, I feel this requires more thought/reviews before proceeding with this. I'm unsure how I feel with this. Could you create an issue so we can chat about the api design regarding a "create react app"  sort of wrapper for doc-kit?

packages/core/src/utils/bootstrap.mjs
avivkeller: Yes and no. Yes, it's a CRA, but it adapts to already set up documentation, rather than creating a whole new app with a package.json, etc

avivkeller:
@ovflowd done, but what are your concerns?

ovflowd:
> @ovflowd done, but what are your concerns?

If doing this is the right path. Even React abandoned housing something like CRA. I get that some tools, such as Next.js themselves have also a way to bootstrap a Next.js app, but I'd argue this shouldn't be part of doc-kit itself, nor we should rewrite our commands to acommodate that. It _should_ probably be its own package, within the monorepo (maybe?)

Plus the PR is introducing 3 different commands:

- serve feels like unnecessary, we don't need to have serve by ourselves imo, doc-kit is not a server, shouldn't try to be at this stage. Defer to users how they want to serve, can be as easy as `npx serve`
- bootstrap is complex needs proper discussion on how it works
- install feels like it's not something doc-kit should do, only package managers. I'd argue that the bootstrap command could allow you to select with generators to install. But in general we shouldn't have a helper for that. Think of unified, they have hundreds of packages. It's often easy to find what you need by looking at docs. AIs can also help.

avivkeller:
> serve feels like unnecessary, we don't need to have serve by ourselves imo, doc-kit is not a server, shouldn't try to be at this stage. Defer to users how they want to serve, can be as easy as npx serve

Serve does not than just serve, it also watches for changes and reports on errors, so users can build + watch their docs, such as `docusaurus start`.

avivkeller:
> install feels like it's not something doc-kit should do, only package managers. I'd argue that the bootstrap command could allow you to select with generators to install. But in general we shouldn't have a helper for that. Think of unified, they have hundreds of packages. It's often easy to find what you need by looking at docs. AIs can also help.

I'm thinking about users who know the name of the generator they want, but not which package it is in. It's fairly simple for us to resolve that by adding a basic downloading command. Bootstrap does this, fwiw.

ovflowd:
> > serve feels like unnecessary, we don't need to have serve by ourselves imo, doc-kit is not a server, shouldn't try to be at this stage. Defer to users how they want to serve, can be as easy as npx serve
> 
> Serve does not than just serve, it also watches for changes and reports on errors, so users can build + watch their docs, such as `docusaurus start`.

Again, unsure if this is needed/should be owned by us. Would appreciate consensus first :see_no_evil:

ovflowd:
> > install feels like it's not something doc-kit should do, only package managers. I'd argue that the bootstrap command could allow you to select with generators to install. But in general we shouldn't have a helper for that. Think of unified, they have hundreds of packages. It's often easy to find what you need by looking at docs. AIs can also help.
> 
> I'm thinking about users who know the name of the generator they want, but not which package it is in. It's fairly simple for us to resolve that by adding a basic downloading command. Bootstrap does this, fwiw.

That's fair, but can't they just go to the website, open searchbar and do that? Why a whole command for that?

bmuenzenmeyer:
the entirety of a getting started experience currently fits "above the fold" within https://doc-kit.nodejs.org/getting-started.html - you can go from empty project to something rendered in seconds. 

i agree that serve/watch might be the most useful. installation i usually prefer to be ergonomic to package managers. users wants to install things in this security environment

my two cents. i probably didnt think through it as deeply as i should have when you asked in Slack Aviv. 

this body of work is less important to me than making the defaults generic. 

sample size is small. getting more feedback is a good idea

avivkeller:
If I broke out the commands into idiomatic PRs, and the docs into this PR, would that be able to land the better docs before render?

PR #1009, opened 2026-08-06
chore: disable install scripts via allowScripts and fix audit vulnerabilities

btea:
## Description
refer to https://github.com/nodejs/doc-kit/pull/997#issuecomment-5186717190

### Check List

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3698,16 +3698,16 @@
       "license": "ISC"
     },
     "node_modules/brace-expansion": {
-      "version": "5.0.7",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.7.tgz",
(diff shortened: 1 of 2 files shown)


MattIPv4 (commented):
The changes here seem fine, but as a more general note, I like to leave any changes to dependencies to Dependabot (or rarely, collaborators) -- having external contributors touching lock files introduces a lot of risk for something being snuck in maliciously.

Again, the diff here is fine, the lockfile change is legit, just a general callout.

avivkeller (approved):

ovflowd (changes requested):
I don't really get why these specific packages to be unallowed.

avivkeller:
@ovflowd It's all of our packges that use install scripts, since we don't need them

PR #1010, opened 2026-08-07
feat(caching): support incremental builds

avivkeller:


packages/core/bin/commands/generate.mjs
```diff
@@ -1,71 +1,16 @@
...
- * @property {string} version
- * @property {string} changelog
- * @property {string} gitRef
- * @property {string} index
- * @property {boolean} minify
- * @property {string} typeMap
- */
-
-export default new Command('generate')
+export default insertCommonOptions(new Command('generate'))
```
ovflowd: new Command('generate') should be assigned to const.

insertCommonOptions should probably be addCommonOptions

avivkeller: Why? We haven't assigned to const before, and IMO a default export is cleaner

ovflowd: We can have a default export, what Im saying is that the part of instantiating Command and passing addCommonOptions should be in different lines.

avivkeller: What if I changed `insertCommonOptions(new Command('generate'))` to `createCommonCommand('generate')`?

ovflowd: Works for me.

packages/core/bin/commands/watch.mjs
```diff
@@ -0,0 +1,149 @@
...
+  }
+};
+
+/**
+ * Debounces rebuilds, and keeps them serialized
+ *
+ * @param {() => Promise<void>} task - The work to run
+ * @returns {{schedule: () => void, cancel: () => void}}
+ */
+const createScheduler = task => {
```
ovflowd, line 50: imo: this should live on a dedicated scheduler utility file

packages/core/bin/utils.mjs
```diff
@@ -16,3 +41,59 @@ export const errorWrap =
...
   };
+
+/**
+ * Creates a command carrying the options shared by every command that resolves
+ * a configuration and runs generators.
+ *
+ * @param {string} name - The command name
+ * @returns {import('commander').Command} The command, for chaining
+ */
+export const createCommonCommand = name =>
```
ovflowd, line 52: nit: probably live on a deidcated command uitils file? avoid bloating utils.mjs with all sort of unrelated different kind of utils?

packages/core/src/cache/hash.mjs
```diff
@@ -0,0 +1,54 @@
+'use strict';
```
ovflowd, line 1: I wonder if we're reinventing the wheel here, can't we use built-in tools to Node?

packages/core/src/cache/index.mjs
```diff
@@ -0,0 +1,344 @@
...
+ * `node_modules/.cache/doc-kit` when a `node_modules` exists, falling back
+ * to `.doc-kit-cache`.
+ *
+ * @param {string} [explicit] - Configured directory, if any
+ * @returns {string} Absolute cache directory
+ */
+export const resolveCacheDir = explicit =>
+  resolve(
+    explicit ||
+      (existsSync('node_modules')
```
ovflowd, line 41: q: will this always correctly resolve to the right node_modules folder? From which context is existsSync called from?

packages/core/src/cache/index.mjs
```diff
@@ -0,0 +1,344 @@
...
+  if (entries.length > 0 && !outputDir) {
+    return false;
+  }
+
+  const checks = await Promise.all(
+    entries.map(async ([rel, record]) => {
+      try {
+        const path = join(outputDir, rel);
+
+        if ((await stat(path)).size !== record.size) {
```
ovflowd, line 67: nit: separate await from statement to declaration

packages/core/src/cache/index.mjs
```diff
@@ -0,0 +1,344 @@
...
+    return (context = null);
+  }
+
+  try {
+    const dir = resolveCacheDir(settings.dir);
+    const outputDir = configuration.global?.output;
+    const store = createStore(dir);
+    const manifest = await loadManifest(dir);
+
+    await resolveTypeMap(configuration);
```
ovflowd, line 124: q: why we need specific handling just for the type map?

packages/core/src/cache/index.mjs
ovflowd, line 1: q: I really enjoy the code in this file, but just wondering if it makes sense for us to create and host a whole cache system.

avivkeller: I don't know of a dependency that does this for us?

packages/core/src/cache/index.mjs
avivkeller: The type map can either be an object or a file path, so we need to resolve it

packages/core/src/cache/index.mjs
avivkeller: `existsSync('node_modules')` will find the `node_modules` folder of the caller. I can replace this with a dependency, but not sure if it's worth the dependency install for ~10 lines of code

packages/core/src/cache/hash.mjs
avivkeller: This is a wrapper around `import { hash } from 'node:crypto';`, so we are using built in tools.

packages/core/src/cache/index.mjs
ovflowd: No idea either, just thinking out loud.

packages/core/src/cache/manifest.mjs
ovflowd, line 1: q: as amount of files increase, how much of time and cpu are we wasting by reading all these files, computing salts/hashes and comparing them against current ones in the pipeline? Are the savings of incremental builds that great?

packages/core/src/cache/salt.mjs
```diff
@@ -0,0 +1,93 @@
...
+
+import { combine, hashValue } from './hash.mjs';
+
+/**
+ * The only code-identity salt. Bump it in any release whose generated output
+ * differs (and whenever the cache's own semantics change) to invalidate every
+ * existing entry; releases that don't change output leave caches valid. When
+ * hacking on generator code locally, build with `--force`.
+ */
+export const CACHE_SCHEMA = 1;
```
ovflowd, line 11: nit: should this be somewhere else?

packages/core/src/cache/manifest.mjs
avivkeller: > Are the savings of incremental builds that great?

Incremental builds allow us to host a dev server, and allow re-builds to take ~10s rather than ~40s+. If we have no intention of this functionality, then it's not really needed

packages/core/src/cache/store.mjs
```diff
@@ -0,0 +1,164 @@
...
+     */
+    prune: async maxAgeDays => {
+      const cutoff = Date.now() - maxAgeDays * 24 * 60 * 60 * 1000;
+
+      const entries = await readdir(objectsDir, {
+        recursive: true,
+        withFileTypes: true,
+      }).catch(() => []);
+
+      await Promise.all(
```
ovflowd, line 147: This is a little bit wacky, wondering if there's a cleaner way of doing this? Also I wonder if cache cleanup should be the job of doc-kit itself... What other tools such as turborepo/next.js do? Do they clean their own cache? And how? Can you compare impls?

packages/core/src/cache/types.d.ts
ovflowd, line 1: I really can't imagine that an existing cache system for the exact sort of thing we're doing doesn't exist. So many frameworks/tools out there need to do the _exact_ same thing. Could you please do me a favor, maybe use Claude, and see what's out there? I really would appreciate not needing to maintain all of this extra code.

packages/react/src/sitemap/generate.mjs
```diff
@@ -49,7 +51,10 @@ export async function generate(entries) {
 
   const urlset = apiPages
     .map(page =>
-      entryTemplate
+      (page.lastmod
```
ovflowd, line 54: note: is this relevant? it feels like it could be done on its own? (separate PR)

ovflowd (changes requested):
This PR is doing so much (watch commands, changes on generators that seem unrelated + creation of a whole caching system) although I like the idea of incremental build supports, how many seconds of improvements are we talking about (assume only fs.md and http.md) got touched?

I want to understand what benefits this is bringing, and if we can decouple the PR a tiny bit. It's really hard to appropriately review such a big PR.

PR #1011, opened 2026-08-07
chore: rename packages

avivkeller:
cc @ovflowd, as discussed on Slack. Breaks out `@doc-kit/cli`, and renames other packages


bmuenzenmeyer (approved):

ovflowd (approved):

CONTRIBUTING.md
```diff
@@ -77,10 +77,11 @@ This repository is an npm workspaces monorepo. The root package is private and
 holds the shared tooling (linting, formatting, tests, changesets); every
 published package lives under `packages/`:
 
-- `packages/core`: [`@nodejs/doc-kit`](packages/core) — the doc-kit engine and CLI
-- `packages/legacy`: [`@nodejs/doc-kit-generator-legacy`](packages/legacy) — the legacy-format generators
+- `packages/cli`: [`@doc-kit/cli`](packages/cli) — the doc-kit command-line interface
+- `packages/core`: [`@doc-kit/core`](packages/core) — the doc-kit engine
+- `packages/legacy`: [`@doc-kit/generator-legacy`](packages/legacy) — the legacy-format generators
```
ovflowd: imo @avivkeller this should not be under @doc-kit but under @node-core, since this is only for Node.js

avivkeller: I put them separately merely for easy deprecation. We can easily deprecate the `legacy` package without deprecating the entire `node` package, but if you disagree I can move it

AugustinMauroy (approved):

PR #1014, opened 2026-08-09
fix(react): keep sidebar tooltips visible

### LINKED ISSUE #938: The tooltips for methods or attributes in the sidebar are not fully displayed.

Image 

Image

As shown in the figure, when the mouse hovers over the `E` mark, the floating information content is covered and not fully displayed.

Similarly, a similar issue occurs when the `E` tag is aligned to the right. It appears to be a problem with the element's layout hierarchy style.

I've encountered a similar issue before; it's likely caused by the container element having a style that hides beyond its horizontal width. It seems like you need to adjust the style or set the tooltip to dynamically adjust its position, moving it closer to the center of the content area.

aswanth-07:
## Summary

- render each stability tooltip in a portal outside the MetaBar scroll container
- keep the tooltip anchored to its badge and clamp it to the viewport
- reposition open tooltips on nested scrolling and viewport resize
- add browser coverage for portal placement, viewport bounds, hover, and keyboard focus

## Why

The table of contents `<dl>` must remain vertically scrollable, but that makes it a clipping ancestor for CSS pseudo-element tooltips. Moving the anchor to the full heading row avoided the clipping while producing the odd alignment noted in review.

The stability badge is now a small hydrated island. Its tooltip renders under `document.body`, so it is no longer a descendant of the scrolling `<dl>` and remains aligned with the badge.

Fixes #938.

## Validation

- `npm run format:check`: passed
- `npm run lint`: passed with three pre-existing warnings and no errors
- focused HTML generator tests: 5/5 passed
- full Playwright suite: 5/5 passed
- full Windows unit run: 537/541 passed; four unrelated assertions expect POSIX path separators and fail on Windows

### DIFF
--- a/.changeset/calm-tooltips-float.md
+++ b/.changeset/calm-tooltips-float.md
@@ -1,0 +1,6 @@
+---
+'@doc-kit/generator-react': patch
+---
+
+Render sidebar stability tooltips in a portal so that the table of contents
+scroll container does not clip them.
--- a/e2e/stability-tooltip.spec.js
+++ b/e2e/stability-tooltip.spec.js
@@ -1,0 +1,33 @@
+import { expect, test } from '@playwright/test';
+
+test('renders stability tooltips outside the MetaBar scroll container', async ({
+  page,
+}) => {
+  await page.goto('/assert.html');
+
+  const badge = page.getByLabel('Stability: Legacy').first();
+  await badge.scrollIntoViewIfNeeded();
+  await expect(badge.locator('xpath=ancestor::is-land[1]')).toHaveAttribute(
+    'ready',
+    ''
+  );
+  await badge.hover();
+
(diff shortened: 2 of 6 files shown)


avivkeller:
Screenshot 2026-08-09 at 7 50 02 AM
This feels oddly aligned, wouldn't it make more sense to fix the issue that clips them rather than work around it?

avivkeller:
@aswanth-07 I'm sure there's a CSS only solution to this either here or upstream. One that doesn't clip, and doesn't misalign

aswanth-07:
Thanks for the feedback. I agree the custom portal workaround is too broad and introduces the alignment problem instead of fixing the underlying tooltip/container CSS. I'm closing this version rather than iterating on the wrong approach. A future fix should be a much smaller CSS or upstream ui-components change.

PR #1015, opened 2026-08-09
chore: move to pnpm

avivkeller:
Now that we are a monorepo, it would be nice to use pnpm (so that `workspace:` imports are rewritten during publish).

I did some source code checking, and we don't actually need to follow the instructions at https://pnpm.io/using-changesets, since [changesets uses the workspace defined tool internally](https://github.com/changesets/changesets/blob/maintenance/v2/packages/cli/src/commands/publish/npm-utils.ts#L84).

Consumers (like node core) can still use npm to install the package.


AugustinMauroy (approved):

ovflowd:
> Now that we are a monorepo, it would be nice to use pnpm (so that workspace: imports are rewritten during publish).

I suppose using doc-kit from Node-core that is `npm`-only will still work? I know that's a dumb question, but just asking in a precautionary fashion.

avivkeller:
Yep :-) It should work fine

ovflowd (approved):
SGTM 🫡

avivkeller:
Requesting fast track as this needs to land before RenderATL

PR #1017, opened 2026-08-09
fix(vercel): clean URLs

avivkeller:


avivkeller:
(This is a hotfix)

avivkeller:
Vercel's previews were incorrectly configured so page links were broken in doc-kit.nodejs.org before this. We don't want a site with no links xD

PR #1018, opened 2026-08-09
chore: preload fonts

avivkeller:
cc @ovflowd

### DIFF
--- a/packages/react/src/html/bundlers/vite.mjs
+++ b/packages/react/src/html/bundlers/vite.mjs
@@ -256,7 +256,7 @@ export const createViteConfig = ({
           format: 'es',
 
           /**
-           *
+           * Determine the asset names for different files
            */
           assetFileNames: asset =>
             asset.names.some(name => name.endsWith('.woff2'))
--- a/packages/react/src/html/bundlers/vite.mjs
+++ b/packages/react/src/html/bundlers/vite.mjs
@@ -11,6 +11,8 @@ import {
   mergeConfig,
 } from 'vite';
 
+import { FONT_DIRECTORY } from '../constants.mjs';
+
 const VIRTUAL_PREFIX = 'virtual:doc-kit/';
 const RESOLVED_VIRTUAL_PREFIX = '\0doc-kit:';
 
@@ -252,6 +254,18 @@ export const createViteConfig = ({
         output: {
           ...vite.build?.rolldownOptions?.output,
           format: 'es',
+
+          /**
+           *
+           */
+          assetFileNames: asset =>
+            asset.names.some(name => name.endsWith('.woff2'))
+              ? // We need to know where the fonts are to preload
+                // them. Using a dynamic hash would make this
+                // difficult.
+                `${FONT_DIRECTORY}/[name][extname]`
+              : 'assets/[name]-[hash][extname]',
+
           ...(server
             ? {
                 entryFileNames: '[name].mjs',
--- a/packages/react/src/html/constants.mjs
+++ b/packages/react/src/html/constants.mjs
@@ -72,6 +72,20 @@ export const JSX_IMPORTS = {
 };
 
 /**
+ * Where the bundler emits fonts
+ */
+export const FONT_DIRECTORY = 'assets/fonts';
+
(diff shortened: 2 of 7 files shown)


packages/react/src/html/bundlers/vite.mjs
```diff
@@ -252,6 +254,18 @@ export const createViteConfig = ({
         output: {
           ...vite.build?.rolldownOptions?.output,
           format: 'es',
+
+          /**
+           *
```
avivkeller:
```suggestion
           * Determine the asset names for different files
```

packages/react/src/html/bundlers/vite.mjs
```diff
@@ -252,6 +254,18 @@ export const createViteConfig = ({
         output: {
           ...vite.build?.rolldownOptions?.output,
           format: 'es',
+
+          /**
+           * Determine the asset names for different files
+           */
+          assetFileNames: asset =>
```
ovflowd, line 261: just ooc: how Next.js preloads fonts?

packages/react/src/html/constants.mjs
```diff
@@ -71,6 +71,20 @@ export const JSX_IMPORTS = {
...
 
+/**
+ * Where the bundler emits fonts
+ */
+export const FONT_DIRECTORY = 'assets/fonts';
+
+/**
+ * Fonts to preload
+ */
+export const FONTS = [
```
ovflowd, line 82: Can we avoid doing static lists? Maybe a function that we define per font we need that creates the css block and the import and the loading like Next.ja?

avivkeller: If we do it that way, it's slower, since we need to minify the dynamically created HTML. This way we can do one minification pass

ovflowd: Why it would be slower? :thinking:

ovflowd: I mean, maybe you're not visualizing the idea:

```tsx
const [codeToAddToHead, fontFamilySnippet] = generateFontSnippet('open-sans-latin-wght-italic', ['woff']);
```

fgirst argument being the font name, 2nd being the supported formats.

avivkeller: > I mean, maybe you're not visualizing the idea:
> 
> 
> 
> ```tsx
> 
> const [codeToAddToHead, fontFamilySnippet] = generateFontSnippet('open-sans-latin-wght-italic', ['woff']);
> 
> ```
> 
> 
> 
> fgirst argument being the font name, 2nd being the supported formats.

Each font has a different snippet, so that's not an option

ovflowd: Yes, each font has a different snippet. So you append instead of https://github.com/nodejs/doc-kit/pull/1018/changes#diff-71b846ed4741eff9604ce9e99abc324eb6dafbb8724dc337f6eefde00dae282aR10

Youll have

${fontOnePreload}
${fontTwoPreload}

ovflowd: This is genuinely what Next.js does, annd makes maintainability much easier/reduces hard-coded pieces.

avivkeller: I don't want to do it like that. I'd prefer a `preloads` section with a hardcoded font list so that the bundler can decide how it does prelaods.

Doing ${preloadA} ${preloadB} makes the template not very overrideable, should we add or remove fonts

ovflowd: I feel we're at an impass :sweat_smile: -- I feel we should avoid hard-coded list of fonts and allow this to be customizable/used by the end-user rather than the way it is lied out currently.

ovflowd: btw we need to reach consensus before RenderATL :see_no_evil:

avivkeller:
Fast-tracked for RenderATL, we can discuss alternatives in a follow-up, if needed

ovflowd:
> Fast-tracked for RenderATL, we can discuss alternatives in a follow-up, if needed

You shouldn't have done this :/

PR #1019, opened 2026-08-09
chore: new index page

avivkeller:
This version is more mobile-friendly, as it does not rely on tables to function


ovflowd (changes requested):
Unfortunately not a fan at all of this design :(

avivkeller:
Any improvement ideas?

PR #1021, opened 2026-08-10
fix: move tsconfig.json to project root

btea:
The packages/core/tsconfig.json only covered packages/core/src, leaving packages/react and packages/node-legacy without a TypeScript project context. This caused type errors in their types.d.ts files that reference global types (GeneratorMetadata, Generate, ProcessChunk) declared in packages/core/src/generators/types.d.ts.

A root tsconfig.json with include: ["packages/*/src"] ensures all packages share a single compilation context where the global type declarations are visible everywhere.

image

### Check List

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,5 +9,5 @@
     "types": ["node"],
     "skipLibCheck": true
   },
-  "include": ["src"]
+  "include": ["packages/*/src"]
 }


avivkeller:
Fast-tracked for RenderATL

ovflowd:
> Fast-tracked for RenderATL

I don't think RenderATL is a reason for fast-tracking arbitrary PRs... I mean, you could have approved the PR and ask for fast-tracking and we'd give a 👍 but should have at least waited for that imo 😅

PR #1022, opened 2026-08-10
chore: export internal components

avivkeller:
As titled; consumers may want to monkey patch these components


AugustinMauroy (approved):

ovflowd (approved):

ovflowd:
> As titled; consumers may want to monkey patch these components

Monkey-patch or extend?

PR #1023, opened 2026-08-11
fix: split non-parameter items from typed lists in signature tables

btea:
## Description
When a loose markdown list in API docs starts with typed parameters (e.g. `actual`, `expected`, `Returns`) but also contains plain prose bullets (e.g. algorithm complexity notes in `util.diff`), the entire list was treated as a parameter signature table. The non-parameter items had no name or type, causing them to render as empty sections.

Split the list at the first non-parameter item: typed items become the FunctionSignature table, and remaining items render as regular markdown.

Affected pages: util, fs, quic (31 mixed lists across 4 files).

image

### Check List

### DIFF
--- a/.changeset/fix-typed-list-split.md
+++ b/.changeset/fix-typed-list-split.md
@@ -1,0 +1,9 @@
+---
+'doc-kit': patch
+---
+
+Fix empty paragraphs in API docs when a typed parameter list contains trailing non-parameter items
+
+When a loose markdown list in the API docs starts with typed parameters (e.g. `actual`, `expected`, `Returns`) but also contains plain prose bullets (e.g. algorithm complexity notes), the entire list was previously treated as a parameter signature table. The non-parameter items had no name or type, causing them to render as empty `<section>` blocks on the built site.
+
+The fix splits the list at the first non-parameter item: typed items become the `FunctionSignature` table, and the remaining items render as regular markdown content.
--- a/.changeset/fix-typed-list-split.md
+++ b/.changeset/fix-typed-list-split.md
@@ -1,5 +1,6 @@
 ---
-'doc-kit': patch
+'@doc-kit/core': patch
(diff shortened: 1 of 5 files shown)


AugustinMauroy (approved):

packages/core/src/utils/queries/utils.mjs
avivkeller, line 63: This feels duplicative

btea: Yeah, the logic behind the judgments is almost identical.

avivkeller (approved):

.changeset/fix-typed-list-split.md
```diff
@@ -0,0 +1,9 @@
+---
+'doc-kit': patch
```
avivkeller:
```suggestion
'@doc-kit/core': patch
'@doc-kit/generator-react': patch
```

ovflowd:
Asking for fast-tracking @nodejs/web-infra

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -264,11 +264,33 @@ export const processEntry = entry => {
   // Transform typed lists into property tables. Skipped for MDX pages, whose
   // lists are authored prose rather than API type signatures.
   if (!entry.mdx) {
-    visit(
-      entry.content,
-      UNIST.isStronglyTypedList,
-      (node, idx, parent) => (parent.children[idx] = createSignatureTable(node))
-    );
+    visit(entry.content, UNIST.isStronglyTypedList, (node, idx, parent) => {
```
ovflowd, line 267: nit: can we clean this up a bit? (follow-up) make it easier to read and follow the logic.

btea: Any suggestions? Extract the last function?

PR #1036, opened 2026-08-13
feat(react): add stability index

### LINKED ISSUE #1031: Concrete example first, then explanation, finally methods list

For many APIs (but not all), I find the following structure of a **docs page** to be really navigable and intuitive for the understanding of the API itself:
1. A commented, concrete mid-size example
2. Description of the API
3. Formal list of methods 
4. (Optional) Extra info about possible nuances, flags or whatever

P.S.
It's already the 4th suggestion I report today but, given the docs are under construction, I'll take advantage of it :). 
This is more of a design decision, so sorry if isn't a "structured" issue reporting

avivkeller:
This PR populates a stability index to the index page, if it marks itself as needed one.

To demo this, the `about.md` page has been moved to `index.md`:

Screenshot 2026-08-13 at 10 50 35 AM

### DIFF
--- a/packages/react/src/jsx-ast/generate.mjs
+++ b/packages/react/src/jsx-ast/generate.mjs
@@ -3,10 +3,10 @@ import { groupNodesByModule } from '@doc-kit/core/utils/generators.mjs';
 import { jsx, toJs } from 'estree-util-to-js';
 
 import buildContent from './utils/buildContent.mjs';
+import { injectDocumentationIndex } from './utils/documentationIndex.mjs';
 import { getSortedHeadNodes } from './utils/getSortedHeadNodes.mjs';
 import { buildNotFoundPage } from './utils/synthetic/404.mjs';
 import { buildAllPage } from './utils/synthetic/all.mjs';
-import { buildIndexPage } from './utils/synthetic/index.mjs';
 
 /**
  * Builds the `{ head, entries }` page descriptors for all configured synthetic
@@ -21,7 +21,6 @@ const buildSyntheticDescriptors = input => {
 
   return [
     config.generateAllPage && buildAllPage(input),
-    config.generateIndexPage && buildIndexPage(input),
     config.generateNotFoundPage && buildNotFoundPage(),
   ].filter(Boolean);
 };
@@ -60,8 +59,14 @@ export async function processChunk(slicedInput, itemIndices) {
  * @type {import('./types').Generator['generate']}
  */
 export async function* generate(input, worker) {
-  // The synthetic `index` page replaces the Core `index` document.
+  // The `index` page is only generated when an `index` document is part of
+  // the input; the module list for the synthetic pages and the stability
+  // overview excludes it.
   const moduleInput = input.filter(entry => entry.api !== 'index');
+
+  // Sections tagged with a `<!-- DOCUMENTATION_INDEX -->` comment (e.g. in
+  // the `index` document) receive the Stability Overview of all modules.
+  injectDocumentationIndex(input, moduleInput);
 
   // Create sliced input: each item contains head + its module's entries
   // This avoids sending all 4700+ entries to every worker
--- a/scripts/vercel-prepare.sh
+++ b/scripts/vercel-prepare.sh
@@ -26,6 +26,8 @@ cd node
 # Enable sparse checkout and specify the folder
 git sparse-checkout set lib doc .
 
+sed 's/STABILITY_OVERVIEW_SLOT_BEGIN/DOCUMENTATION_INDEX/g' ./doc/api/documentation.md > ./doc/api/index.md
+
 # Move back out
 cd ..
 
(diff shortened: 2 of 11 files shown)


AugustinMauroy (changes requested):
https://api-docs-tooling-git-stability-index-openjs.vercel.app/documentation#stability-overview

the page is still build but not completely populate.
IMO we shouldn't build it from node core source but duplicate the index page as documentation for now and in next node core major remove this page

avivkeller:
> IMO we shouldn't build it from node core source but duplicate the index page as documentation for now and in next node core major remove this page

What? Different projects may have different abouts, so we (IMO) 100% should not hardcode this "Node.js"-themed about.

> the page is still build but not completely populate.

https://api-docs-tooling-git-stability-index-openjs.vercel.app/ is the correct URL for that, fixed, thanks!

scripts/vercel-prepare.sh
```diff
@@ -26,6 +26,9 @@ cd node
 # Enable sparse checkout and specify the folder
 git sparse-checkout set lib doc .
 
+sed 's/STABILITY_OVERVIEW_SLOT_BEGIN/DOCUMENTATION_INDEX/g' ./doc/api/documentation.md > ./doc/api/index.md
+rm ./doc/api/documentation.md
```
AugustinMauroy, line 30: Note that this is a content change not design/build pipeline.  IMO we must not change the content for now but only change the ui/way to have info.

So to solve that just copy the index.html as documentation.md

AugustinMauroy (approved):
in term of UI/UX that super cool ! but removing page `documentation` not seem good idea to do it now and like that.

scripts/vercel-prepare.sh
avivkeller: The plan (at least for me) is to duplicate this change in core as a part of the migration, so making the preview mirror that ideal change is something I wanted to do. We spoke a bit on this in a web team meeting, and iirc we agreed that moving these pages was ideal, hence this change

avivkeller:
@nodejs/web-infra

MattIPv4:
I'm supportive of bringing this content onto the index page. Having more informational content on the landing page about how the docs work and how someone should consume them seems like a sensible thing.

However, I still think we should do far more to make the about page far more usable to folks new to Node.js trying to find something in the docs: perhaps a full-width hero search box, perhaps cards for the different docs pages instead of the table, and perhaps placing those cards before the information about the stability levels?

None of that has to block this PR though; I think this is a great first step to making that index/landing page more useful.

avivkeller:
Noted! Would this be good for this PR simply moving the content over, and then a follow-up making it nice and pretty with cards?

MattIPv4:
> Noted! Would this be good for this PR simply moving the content over, and then a follow-up making it nice and pretty with cards?

Yep, my comment was non-blocking (not approving only because I don't know this codebase well enough to confidently do that).

avivkeller:
Once this PR gets landed I'll open a follow-up. I can probably use the same logic we use in llms.txt to make little description cards

scripts/vercel-prepare.sh
```diff
@@ -26,6 +26,9 @@ cd node
 # Enable sparse checkout and specify the folder
 git sparse-checkout set lib doc .
 
+sed 's/STABILITY_OVERVIEW_SLOT_BEGIN/DOCUMENTATION_INDEX/g' ./doc/api/documentation.md > ./doc/api/index.md
```
ovflowd, line 29: nit: add a todo that this is temporary till work on node core is done?

packages/react/src/jsx-ast/generate.mjs
```diff
@@ -60,9 +59,15 @@ export async function processChunk(slicedInput, itemIndices) {
...
 export async function* generate(input, worker) {
-  // The synthetic `index` page replaces the Core `index` document.
+  // The `index` page is only generated when an `index` document is part of
+  // the input; the module list for the synthetic pages and the stability
+  // overview excludes it.
   const moduleInput = input.filter(entry => entry.api !== 'index');
 
+  // Sections tagged with a `<!-- DOCUMENTATION_INDEX -->` comment (e.g. in
+  // the `index` document) receive the Stability Overview of all modules.
+  injectDocumentationIndex(input, moduleInput);
```
ovflowd, line 69: I suppose this is Node.js's specific. I don't know if this should be on jsx-ast. I feel that it should probably be a custom Node.js step for doc-kit. In the future `<!-- DOCUMENTATION_INDEX -->` should completely be removed from source and it should simply be rendered differenty. (ie: a React component or something)

ovflowd (approved):

ovflowd:
image

Can this be properly rendered?

packages/react/src/jsx-ast/generate.mjs
avivkeller: It's not node specific, other doc sites need an index

ovflowd: Let me rephrase, it is not in our specification, so it needs to be updated in a way that is conforming or part to the spec.

avivkeller: Our specification doesn't designate per-generator extensions, which this is. It's documented in the generator dcumentation

PR #1038, opened 2026-08-14
fix(jsx-ast): hide redundant headers and badges for overloaded methods

moshams272:
## Description

This PR removes the heading from the overloads function; just keep one for the first function that appears like grouping them.

### Before:
image

### After:
image

## Related Issues

None

### Check List

- [X] I have run `pnpm test` and all tests passed.
- [X] I have check code formatting with `pnpm run format:check` & `pnpm run lint`.

### DIFF
--- a/packages/react/src/jsx-ast/utils/__tests__/buildContent.test.mjs
+++ b/packages/react/src/jsx-ast/utils/__tests__/buildContent.test.mjs
@@ -3,7 +3,10 @@ import { describe, it } from 'node:test';
 
 import { setConfig } from '@doc-kit/core/utils/configuration/index.mjs';
 
-import { transformHeadingNode } from '../buildContent.mjs';
+import {
+  transformHeadingNode,
+  createHeadingElement,
+} from '../buildContent.mjs';
 
 const heading = {
   type: 'heading',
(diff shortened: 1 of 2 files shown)


avivkeller:
cc @nodejs/web thoughts?

AugustinMauroy:
when you have long type table for the constructor the overlaod table seem alone

moshams272:
> when you have long type table for the constructor the overlaod table seem alone

What if we make a **Nested ToC** like we keep the output of that PR, but make the codeblock, that contains its signature, has an anchor id, and put it under the main one with its signature in the ToC.

Something like that:
```md
- factorizeModule
    - factorizeModule(options, callback): void
    ...
```
But it'll not be good in ToC, WDUT?!

moshams272:
Or creating a table under the header and before the function, that table contains all signature with an anchor id for each of them, so we will have one link in ToC.

ovflowd (approved):
I think I'm fine with this, but wonder if all overloads should simply be added to one signature. (the table gets redesigned/updated) to ensure the Type box and the table of properties/return type follow the possible overloads. But that's a more complex PR that'd probably require more design.

I can even imagine that using code tabs could be useful for overloads. So instead of rendering each overload below each other, we have tabs, default being first implementation, then the other tabs being overloads. What do you all think?

moshams272:
> I can even imagine that using code tabs could be useful for overloads. So instead of rendering each overload below each other, we have tabs, default being first implementation, then the other tabs being overloads. What do you all think?

Need to put +1, but I'm not member, so... +0.1 ;)

PR #1040, opened 2026-08-15
fix: export MetadataEntry types from @doc-kit/core

btea:
## Description
Add explicit exports map entries for the metadata types subpaths (`./metadata/types` and `./generators/metadata/types`) pointing to `types.d.ts`. Previously, these paths only resolved through the `./*` wildcard, which fails under bundler module resolution since an extensionless specifier can't resolve to a `.d.ts` file. This caused JSDoc `import('@doc-kit/core/generators/metadata/types')` type annotations across the react/node packages to silently fail, leaving params typed as `any`.

image

### Check List

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,6 +21,14 @@
     "./ast-js": "./src/generators/ast-js/index.mjs",
     "./json-simple": "./src/generators/json-simple/index.mjs",
     "./metadata": "./src/generators/metadata/index.mjs",
+    "./metadata/types": {
+      "types": "./src/generators/metadata/types.d.ts",
+      "default": "./src/generators/metadata/types.d.ts"
+    },
+    "./generators/metadata/types": {
+      "types": "./src/generators/metadata/types.d.ts",
+      "default": "./src/generators/metadata/types.d.ts"
+    },
     "./package.json": "./package.json",
     "./shiki.config.mjs": "./shiki.config.mjs",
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):

packages/core/package.json
```diff
@@ -21,6 +21,14 @@
...
+      "types": "./src/generators/metadata/types.d.ts",
+      "default": "./src/generators/metadata/types.d.ts"
+    },
+    "./generators/metadata/types": {
+      "types": "./src/generators/metadata/types.d.ts",
+      "default": "./src/generators/metadata/types.d.ts"
+    },
     "./package.json": "./package.json",
     "./shiki.config.mjs": "./shiki.config.mjs",
     "./src/*": "./src/*",
```
avivkeller, line 26: Can't we use an array here instead, then it would be able to infer with extensions?

btea: I don't understand. Could you provide an example?

avivkeller: https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/package.json#L7-L13:

```
[
        "./src/*",
        "./src/*.tsx",
        "./src/*/index.tsx",
        "./src/*.ts",
        "./src/*/index.ts"
      ]
```

btea: It looks feasible. 👍

avivkeller (approved):

PR #1041, opened 2026-08-15
ci: resolve PR number from head repo/branch for fork PRs 

btea:
## Description
The "Leave a comment" workflow fails because `thollander/actions-comment-pull-request` runs on the deprecated Node 20 runtime [1], and the upstream PR upgrading it to Node 24 [2] has been open without activity since March 2026.                          

Replace the action with a local composite action built on actions/github-script. The marker format is kept so existing
tagged comments are updated instead of duplicated.                               
                                                                               
[1] https://github.com/nodejs/doc-kit/actions/runs/31868656146/job/94973630400  
[2] https://github.com/thollander/actions-comment-pull-request/pull/449 

### Check List


MattIPv4 (changes requested):
I don't think Node.js 20/24 has anything to do with that failure -- the workflow was triggered from a commit on main, and so `github.event.workflow_run.pull_requests[0].number` used as the input will not be populated.

btea:
You're right, I didn't address the error; I simply removed the warning message about the Node version. I've rewritten the logic, but it doesn't seem to be directly verifiable in this PR.

avivkeller (changes requested):
I'm also wary to replace something that works, if it ain't broke don't fix it... why do we need our own logic when an external action works fine

btea:
The local logic was added to resolve a Node.js version compatibility warning. The upstream action hasn't been updated in a long time, suggesting the author is no longer managing it. If you believe this warning can be ignored, I can revert the code and continue using the upstream action.

image

MattIPv4:
👍 Yes, no need to introduce additional maintenance burden by rolling our own action, the upstream works w/ Node.js 24.

MattIPv4 (commented):
Can we follow the same pattern as in https://github.com/nodejs/nodejs.org/blob/main/.github/workflows/leave-comment.yml, please, so we have consistent implementations?

btea:
It turns out there's a well-established logic to refer to; thank you for pointing that out.

MattIPv4:
@btea could I ask if you're using AI for your contributions and/or your comments?

btea:
> @btea could I ask if you're using AI for your contributions and/or your comments?

Yes, this PR was written using AI, and I'm not very familiar with Actions. The comments were handwritten. (However, I used a translation tool because my English isn't very good.)

MattIPv4 (approved):

avivkeller (approved):

AugustinMauroy (approved):

PR #1045, opened 2026-08-17
chore: improved index page

avivkeller:
Screenshot 2026-08-17 at 10 48 33 AM


AugustinMauroy:
Capture d’écran 2026-08-17 à 22 24 07


you should also transform description

AugustinMauroy:
Capture d’écran 2026-08-17 à 22 25 56

MattIPv4:
Can we maybe clamp the descriptions to 3 or 4 lines? And perhaps allow the rest to show via overflow (not shifting the grid layout) on hover?

Can we move the cards up to be before the contributing + stability index headings? I think for most folks hitting the docs, the primary action/path is going to be wanting to find a section of the docs?

packages/react/src/html/ui/components/MetaBar/index.jsx
```diff
@@ -25,9 +22,7 @@ const HeadingValue = ({ value, stability }) => {
     return value;
   }
 
-  const ariaLabel = STABILITY_TOOLTIPS[stability]
-    ? `Stability: ${STABILITY_TOOLTIPS[stability]}`
-    : undefined;
+  const label = STABILITY_LABELS[stability];
```
ovflowd, line 25:
```suggestion
  const [label] = STABILITY_LABELS[stability] ?? [];
```

packages/react/src/html/ui/components/MetaBar/index.jsx
```diff
@@ -37,11 +32,11 @@ const HeadingValue = ({ value, stability }) => {
...
         className={styles.badge}
         kind={STABILITY_KINDS[stability]}
-        data-tooltip={STABILITY_TOOLTIPS[stability]}
-        aria-label={ariaLabel}
+        data-tooltip={label}
+        aria-label={label ? `Stability: ${label}` : undefined}
         tabIndex={0}
       >
-        {STABILITY_LABELS[stability]}
+        {label?.[0]}
```
ovflowd, line 39: The suggestion I made on line 25 makes this line better imo

packages/react/src/html/ui/components/constants.mjs
```diff
@@ -0,0 +1,12 @@
+/**
+ * UI badge kinds and labels for Node.js API stability levels
+ *
+ * @see https://nodejs.org/api/documentation.html#stability-index
+ */
+export const STABILITY_KINDS = ['error', 'warning', 'default', 'info'];
+export const STABILITY_LABELS = [
```
ovflowd, line 7: Don't we have already these constants somewhere? Like on @node-core/ui-components?

packages/react/src/html/ui/components/constants.mjs
```diff
@@ -0,0 +1,12 @@
+/**
+ * UI badge kinds and labels for Node.js API stability levels
+ *
+ * @see https://nodejs.org/api/documentation.html#stability-index
+ */
+export const STABILITY_KINDS = ['error', 'warning', 'default', 'info'];
```
ovflowd, line 6: I'd argue this is `STABILITY_TYPES` and I'd probably argue the Labels Array and Kinds should be a Map where the pair key is the Label and the value the Badge Type. Also adding a comment that these = type of UI Badges we have and maybe ensure it follows types that are supported by our Badges would also be good.

packages/react/src/html/constants.mjs
```diff
@@ -26,6 +26,10 @@ export const JSX_IMPORTS = {
     name: 'CodeTabs',
     source: resolve(ROOT, './ui/components/CodeTabs'),
   },
+  DocumentationIndex: {
```
ovflowd, line 29: Is this DocumentationIndex sort of component just for Node, or going to be kinda part of standard doc-kit components? I feel that all the components doc-kit exposes on our web generator should be documented under the web generator docs [...]

I also feel tha this name "DocumentationIndex" doesn't really say what this component renders, can we have a more descriptive name for the component, tgat'd b great

packages/react/src/jsx-ast/constants.mjs
```diff
@@ -189,6 +189,9 @@ export const AST_NODE_TYPES = {
   },
 };
 
+// `<!-- DOCUMENTATION_INDEX -->` comment in a source document = a stability index
```
ovflowd: nit: add a todo that we aim to remove this from node-core, or that it follows our standard doc-kit YAML syntax.

We might need to expand our spec on the supported type of slots, sinca our YAML snippets are "slots" that add metadata to render certain things, we might as well have something that is either just MD-compatible/doc-kit spec compatible or MDX-compatible (even if strictly not MDX)...

ovflowd: Or we outright just say that doc-kit spec extends CommonMark and the MDX spec. 🤷  since we're already using remark [...], recma [...] probably we could just support mdx, and support .mdx extensions to files in case they want to add MDX content.

I think that makes much sense and allows more customization to the pages by downstream consumers of doc-kit.

packages/react/src/jsx-ast/generate.mjs
```diff
@@ -60,13 +83,16 @@ export async function processChunk(slicedInput, itemIndices) {
...
-  // the input; the module list for the synthetic pages and the stability
-  // overview excludes it.
+  // the input; the module list for the synthetic pages and the documentation
+  // index excludes it.
   const moduleInput = input.filter(entry => entry.api !== 'index');
 
-  // Sections tagged with a `<!-- DOCUMENTATION_INDEX -->` comment (e.g. in
-  // the `index` document) receive the Stability Overview of all modules.
-  injectDocumentationIndex(input, moduleInput);
+  // Sections tagged with a `<!-- DOCUMENTATION_INDEX -->` build an index
```
ovflowd: See, this is what I don't like. I hope in the future, based on my [comment above](https://github.com/nodejs/doc-kit/pull/1045/changes#r3804836736) can alllows us to avoid such specific logic for this sort of scenarios 😅

packages/react/src/jsx-ast/generate.mjs
```diff
@@ -1,13 +1,36 @@
...
+import { JSX_IMPORTS } from '../html/constants.mjs';
 import { buildNotFoundPage } from './utils/synthetic/404.mjs';
 import { buildAllPage } from './utils/synthetic/all.mjs';
 
+/**
+ * Builds the `<DocumentationIndex />` element
+ *
+ * @param {Array<import('@doc-kit/core/generators/metadata/types').MetadataEntry>} moduleEntries
+ */
+const buildDocumentationIndex = moduleEntries =>
```
ovflowd: I genuinely feel we gotta call this something else, I'd also argue it shouldn't be the responsibiliy of this generator to take care of this, but the web generator. That said, I'm fine as this as a temporary solution. But we gotta add deprecated and todo marks and open an issue to keep track of remaining changes.

ovflowd (approved):
I'm fine SGTM'ing this, but with big caveats that this is temporary and requires proper changes on the long-term 🙇

packages/react/src/html/constants.mjs
avivkeller: It's a standard doc kit component. The name `DocumentationIndex` is fairly descriptive of what the component does: Documentation Index

packages/react/src/jsx-ast/constants.mjs
avivkeller: We do support MDX. I can probably replace the comment with an MDX component.

packages/react/src/html/constants.mjs
ovflowd: I unfortunately disagree. "Documentation Index" isn't fairly descriptive, and doesn't say what is rendered. What is in this "documentation index"? I feel that a proper name is "StabilityOverview" or "DocumentationModulesOverview" or something more descriptive.

packages/react/src/jsx-ast/constants.mjs
ovflowd: Cool, that can be follow-up as require change on node core.

avivkeller: -1 on a follow-up. I'll do it here, and update our beta script to make that change, and that can be the TODO, rather than an implementation that is going to be replaced based on one consumer's need.

ovflowd: I don't understand. The source YAML annotation is in node core. I don't think that we should do a hack here to go around that. We can proceed as is, and then add the feature here, and then migrate it over there.

avivkeller: I'd rather make a "hack" to replace the `<!-- -->` syntax of Node core with a `<XYZ />` to go around that then write the entire component in a format that will be changed in a follow-up

ovflowd: I understand you'd do that on a bash script, but I'm unsure about the precedent of that, plus that feels more like doing a monkey patch. I'm not particularly against, just saying that it isn't a good thing to do in general.

AugustinMauroy:
Capture d’écran 2026-08-21 à 18 15 04


@avivkeller you just land the pr with broken style ...

avivkeller:
@AugustinMauroy That's because the broken style isn't something that can be fixed in this PR, it depends on a change in UI components (https://github.com/nodejs/nodejs.org/pull/9115) to allow to better style and transform those.

I'm working on a follow up, but I thought it best not to block this PR.

AugustinMauroy:
noice !

avivkeller:
Well, technically I could've fixed it in this PR for the index page only, but ChangeHistory suffers from the same issue, so I figured that the since root cause was unrelated to this PR, it's fine to land.

Although, it's much less noticeable in ChangeHistory since those are collapsed

ovflowd:
image

I suppose these HTML entities not rendering is expected?

PR #1046, opened 2026-08-18
fix(react): avoid clipped stability tooltips

### LINKED ISSUE #938: The tooltips for methods or attributes in the sidebar are not fully displayed.

dabidabi:
## Summary

- use the browser-native `title` tooltip for stability badges in the meta bar
- keep the existing accessible label and keyboard focus behavior
- add a generated-HTML regression test and a patch changeset

## Why

The shared CSS tooltip is rendered inside the meta bar's scrolling table of
contents. When a stability badge wraps near either horizontal edge, that
container clips part of the tooltip. A native tooltip is rendered outside the
scroll container, so its text remains fully visible without adding client-side
JavaScript to the otherwise static table of contents.

Fixes #938.

## Validation

- `prettier --check` on the changed files
- ESLint on the changed JavaScript files
- targeted web generator tests: 5 passed
- full test suite: 561 passed

AI assistance was used to investigate, implement, test, and review this change.
The commit includes an `Assisted-by: OpenAI Codex` attribution.

### DIFF
--- a/.changeset/fix-metabar-stability-tooltip.md
+++ b/.changeset/fix-metabar-stability-tooltip.md
@@ -1,0 +1,6 @@
+---
+'@doc-kit/generator-react': patch
+---
+
+Prevent stability badge tooltips in the meta bar from being clipped by the
+scrolling table of contents.
--- a/packages/react/src/html/__tests__/generate.test.mjs
+++ b/packages/react/src/html/__tests__/generate.test.mjs
@@ -70,9 +70,12 @@ describe('web generate', () => {
     const { output } = await createTestConfiguration(context);
     const fs = createEntry('fs', 'File system');
     fs.path = '/api/fs';
+    const experimental = createEntry('experimental', 'Experimental API');
+    experimental.heading.depth = 2;
+    experimental.stability = { data: { index: '1' } };
     const notFoundPage = buildNotFoundPage();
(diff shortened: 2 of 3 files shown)


AugustinMauroy (commented):
okay it's worked, but it's a regression that was better before I mean the fix should be applied on the tool-tips component

ovflowd (changes requested):
This switches to native browser tooltips whereas the ask is to ensure the current custom tooltips (generated via data-tooltip) don't have the issue. i.e.: this should be as simple as fixing the source CSS behind the tooltips. That might be on @node-core/ui-components or on doc-kit itself.

btea:
Using the native title seems to avoid the problem, but it loses the relevant styles and doesn't look very aesthetically pleasing. However, it seems that we still need to use JavaScript to maintain the appearance.

PR #1047, opened 2026-08-19
feat(ui): group overloaded functions into tabs

moshams272:
## Description

This PR introduces a new tabbed UI for overloaded functions. This implementation is a direct result of the discussions in #1038. It updates the MDX AST to group overloads into `CodeTabs` component and remove duplicate signatures.

_**Note:** Using AI to write JSDocs & traverse some errors ;)_

## Validation

Add tests, and there are visual results:

**Before:**

image

image

**After:**

image

image

## Related Issues

None

### Check List

- [X] I have check code formatting with `node --run format:check` & `node --run lint`.


moshams272:
You can view the changes from here: [vercel-preview](https://api-docs-tooling-git-fork-moshams272-feat-group-o-f79d4a-openjs.vercel.app/buffer#static-method-bufferfromarray)

avivkeller:
It's an interesting idea, but I think it makes searching harder, e.g. if I want to search for `Buffer.from(buffer): Buffer`, I need to make sure I'm on the right tab

avivkeller:
Oddly enough, I'm fine with the showing of multiple overloads in our current state, and consumers can reformat their docs how they like.

Thanks for this demo, however :-)!

moshams272:
Hmm, fair enough! Thanks for your time 🤍

ovflowd:
I actually have a different opinion and was thinking of this differently. Instead of the simple 1, 2 on the Overload headers which makes distinction hard, I feel that it could look like Code Tabs; So it becomes pretty apparent that there are more versions of this method.

What it means is that the main method signature is always the same (the initial tiny snippet box under the header) and only the attributes and descriptions/etc are under "codetabs"

I still feel that this would be much simpler.

> and consumers can reformat their docs how they like.

Well I do feel our web generator could have a default stance on overloads, as it is already an opinionated style.

ovflowd:
> It's an interesting idea, but I think it makes searching harder, e.g. if I want to search for `Buffer.from(buffer): Buffer`, I need to make sure I'm on the right tab

That much I agree. I think that anchorinhg can be solved with the tab being switched to the right one based on the click... For pure command+f/ctrl+f searches, I'm unsure of either, we could either think a bit more about this or keep the current status quo.

My goal was to make overloads simpler to identify / consume less vertical space and be better organized.

moshams272:
I initially thought about using the signatures/parameters as the tab titles, but I hit two major edge cases:

- Extremely long parameters: If an overload has 4-5 parameters or long variable names, the tab title becomes massive and completely breaks the UI. (AugustinMauroy actually pointed out a similar [issue](https://github.com/nodejs/doc-kit/pull/1038#issuecomment-5302908810)).

- Identical signature, different types: If we only show the signature, the tabs will look identical!

That's why I retreated to indexes, and to make clear that tabs are for overloads, I've placed a header before the component itself (above indexes) contains "Overloads".

After Aviv's comment, I searched about it and found that the web standard solution is using 
Accordion(details & summary tags) , It'll solve the search problem but it'll be vertical, not horizontal. we can put the signature also, but I imagined that design and I hate it, so I closed this PR ;)

ovflowd:
> Extremely long parameters: If an overload has 4-5 parameters or long variable names, the tab title becomes massive and completely breaks the UI. (AugustinMauroy actually pointed out a similar https://github.com/nodejs/doc-kit/pull/1038#issuecomment-5302908810).

Could you show an example of what you're thinking? Are you referring to header sections?

> Identical signature, different types: If we only show the signature, the tabs will look identical!

Im unsure I got what you're saying. Im saying that the initial signature box should be shared across overloads and the only tabbed pieces are the attributes/descriptions/examples

image

So this is not tabbed, but the below gets tabbed

image

And the tabbing design could be that the whole thing gets surrounded by the same box/tabbing header we use for code tabs, but instead of the content being code, it is the actual content of that overload (the attributes table, description, example, code boxes...) and the title of each tab is "Overload #1", "Overload #2"

moshams272:
> Could you show an example of what you're thinking? Are you referring to header sections?
Imagne that the function signature has 10 params...  

That header "Overloads" tells the developers that they r overloads:
Screenshot from 2026-08-19 23-52-12

you can review the "After" image in description of that PR as well.

> Im unsure I got what you're saying. Im saying that the initial signature box should be shared across overloads and the only tabbed pieces are the attributes/descriptions/examples
> And the tabbing design could be that the whole thing gets surrounded by the same box/tabbing header we use for code tabs, but instead of the content being code, it is the actual content of that overload (the attributes table, description, example, code boxes...) and the title of each tab is "Overload #1", "Overload #2"

Overloads siganture not always the same to do that 👀, For example:

```js
lol(option, callback)
lol(index, option, callback)
```

ovflowd:
> Overloads siganture not always the same to do that 👀, For example:
> 
> ```js
> lol(option, callback)
> lol(index, option, callback)
> ```

Fair, but since we show a typescript-representable version of a type, shouldn't then we also render it in a way that renders overloads? Like instead of one line, you could argue it has two lines then with the two overloads? Just trying to think out loud of ways to represent this. You could maybe throw the problem at Claude to deep search how other docs do overloads and see different examples?

moshams272:
I actually saw that design before in mdn, lemme search...

moshams272:
Oh here it is: [MDN/Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#syntax), Did u mean like that?

plus this:
> ...and the title of each tab is "Overload #1", "Overload #2"

I'm +1

ovflowd:
> Oh here it is: [MDN/Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#syntax), Did u mean like that?
> 
> plus this:
> 
> > ...and the title of each tab is "Overload #1", "Overload #2"
> 
> I'm +1

Can you give it a try?

moshams272:
WIP, with new PR 👀

Or I should do it in this PR?!

moshams272:
I updated the result image in PR description, PTAL :)

ovflowd:
@moshams272 could you actually use Code Tabs?

The idea is that each one of these Overload #1, Overload #2 are actual tabs, and the content itself is surrounded as a box

image

Imagine this, but instead of code inside it is the actual contents of that overload (so description, attribute box, examples, etc...)

Can you do that? It will make the UX much better IMO, please rebase PR too!

ovflowd:
Also the signature box, shouldn't have inline comment of `// Overload 1`, ... `// Overload 2` since MDN also doesn't do that.

moshams272:
> Also the signature box, shouldn't have inline comment of `// Overload 1`, ... `// Overload 2` since MDN also doesn't do that.

I made it for readability, as it's easy to know which  overload tab I need to open instead counting the signature by myself to reach the attributes that I need, WDUT :)

moshams272:
> @moshams272 could you actually use Code Tabs?
> 
> The idea is that each one of these Overload #1, Overload #2 are actual tabs, and the content itself is surrounded as a box
> 
> image
> Imagine this, but instead of code inside it is the actual contents of that overload (so description, attribute box, examples, etc...)
> 
> Can you do that? It will make the UX much better IMO, please rebase PR too!

I'll see if that's applicable.

packages/react/src/html/ui/index.css
```diff
@@ -78,6 +78,33 @@ main {
...
+    border: 1px solid var(--color-neutral-200);
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+    border-bottom-left-radius: 0.25rem;
+    border-bottom-right-radius: 0.25rem;
+    padding: calc(var(--spacing) * 4);
+    padding-bottom: calc(var(--spacing) * 2);
+  }
+
+  @media (min-width: 48rem) {
```
ovflowd, line 95: Out of curiosity, are these all the gap overrides we have? I believe that there were gap settings for mobile, tablet, ... [...] (different sizes)

moshams272: yeah, but I saw that enough for gap & padding. I actually need to remove gap from `@madia` 😅

ovflowd: What I meant is are these all the @media tags we need? Or do we need more?

moshams272: No I think that's good.

ovflowd: So you're confirming there is no other @media tag needed here? No other breakpoint for other screen sizes?

moshams272: Yeah, I confirm that no `@media` tag is needed.

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
+    });
+  };
+
+  /**
+   * Extracts the raw signature string from an API entry node and removes the signature node from its children.
+   * @param {import('estree').Node} node - The AST node representing the API entry.
+   * @returns {string|null} The raw TypeScript signature string, or null if not found.
+   */
+  const extractSignature = node => {
+    const sigIdx = (node.children || []).findIndex(
```
ovflowd: nit:

```suggestion
    const signatureIndex = (node.children || []).findIndex(
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
 
+/**
+ * Groups consecutive overloaded function API entries into a single OverloadTabs component.
+ * @param {Array<import('estree').Node>} processedChildren - The processed JSX AST nodes for the API entries
+ * @param {Array<import('@doc-kit/core/generators/metadata/types').MetadataEntry>} originalEntries - The original API metadata entries containing the overload flags
+ * @returns {Array<import('estree').Node>} The final array of layout children with overloads grouped
+ */
+export const groupOverloadsIntoTabs = (processedChildren, originalEntries) => {
+  const finalChildren = [];
+  let activeOverloadGroup = null;
```
ovflowd:
```suggestion
  let activeOverloadGroup;
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
+   * @returns {string|null} The raw TypeScript signature string, or null if not found.
+   */
+  const extractSignature = node => {
+    const sigIdx = (node.children || []).findIndex(
+      c =>
+        c.properties?.className?.includes('signature') ||
+        c.properties?.class === 'signature'
+    );
+    if (sigIdx !== -1) {
+      const sigNode = node.children.splice(sigIdx, 1)[0];
```
ovflowd: nit:


```suggestion
      const signatureNode = node.children.splice(sigIdx, 1)[0];
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
+      return sigNode.properties?.dataSignatureRaw;
+    }
+    return null;
```
ovflowd: nit: default to return nothing, so it is undefined.

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
+  const processOverloadNode = node => {
+    const sigRaw = extractSignature(node);
+    if (sigRaw) {
+      activeOverloadGroup.signatures.push(sigRaw);
+    }
+    activeOverloadGroup.tabsNode.children.push(wrapInDiv(node));
+  };
+
+  for (const [i, current] of processedChildren.entries()) {
+    if (originalEntries[i].heading?.data?.isOverload) {
```
ovflowd: Can we simply the if statements or make early return/continue; logic?

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,140 @@ export const processEntry = entry => {
...
+    activeOverloadGroup = null;
+  };
+
+  /**
+   * Processes a single API entry node belonging to an overload group.
+   * It extracts its signature and pushes its remaining content into a new tab panel.
+   * @param {import('estree').Node} node - The AST node to process and add to the active group.
+   */
+  const processOverloadNode = node => {
+    const sigRaw = extractSignature(node);
```
ovflowd: feels like this can be a ternary

ovflowd (approved):
SGTM! Left a final round of reviews. Would love to see @nodejs/web-infra also giving their opinions on the preview/final look so far.

AugustinMauroy (approved):
Love this !

packages/react/src/jsx-ast/utils/buildContent.mjs
moshams272: For me, no. Could you add a suggestion?

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
+      c =>
+        c.properties?.className?.includes('signature') ||
+        c.properties?.class === 'signature'
+    );
+
+    if (signatureIndex !== -1) {
+      const signatureNode = node.children.splice(signatureIndex, 1)[0];
+      return signatureNode.properties?.dataSignatureRaw;
+    }
+    return;
```
ovflowd: @moshams272 tiny nit, no need for this dangling return operator. the fn will reutnr undefined by default

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
+   */
+  const extractSignature = node => {
+    const signatureIndex = (node.children || []).findIndex(
+      c =>
+        c.properties?.className?.includes('signature') ||
+        c.properties?.class === 'signature'
+    );
+
+    if (signatureIndex !== -1) {
+      const signatureNode = node.children.splice(signatureIndex, 1)[0];
```
ovflowd: nit: cleaner and easie to read

```suggestion
      const [signatureNode] = node.children.splice(signatureIndex, 1) ?? [];
      
      return signatureNode.properties?.dataSignatureRaw;
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
+      children: rootNode.children || [],
+    });
+  };
+
+  /**
+   * Extracts the raw signature string from an API entry node and removes the signature node from its children.
+   * @param {import('estree').Node} node - The AST node representing the API entry.
+   * @returns {string|undefined} The raw TypeScript signature string, or undefined if not found.
+   */
+  const extractSignature = node => {
```
ovflowd: nit: easier to read, no need of `(node.children || [])`

```suggestion
  const extractSignature = ({ children = [] }) => {
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
 
+/**
+ * Groups consecutive overloaded function API entries into a single OverloadTabs component.
+ * @param {Array<import('estree').Node>} processedChildren - The processed JSX AST nodes for the API entries
+ * @param {Array<import('@doc-kit/core/generators/metadata/types').MetadataEntry>} originalEntries - The original API metadata entries containing the overload flags
+ * @returns {Array<import('estree').Node>} The final array of layout children with overloads grouped
+ */
+export const groupOverloadsIntoTabs = (processedChildren, originalEntries) => {
+  const finalChildren = [];
+  let activeOverloadGroup;
```
ovflowd, line 329: I wonder how this handles in asynchronous work? no issues here, right?

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
+
+    activeOverloadGroup = undefined;
+  };
+
+  /**
+   * Processes a single API entry node belonging to an overload group.
+   * It extracts its signature and pushes its remaining content into a new tab panel.
+   * @param {import('estree').Node} node - The AST node to process and add to the active group.
+   */
+  const processOverloadNode = node => {
```
ovflowd, line 413: nit:

```tsx
const processOverloadNode = node => {
  const signatureRaw = extractSignature(node);
  
  signatureRaw && activeOverloadGroup.signatures.push(signatureRaw);
  activeOverloadGroup.tabsNode.children.push(wrapInDiv(node));
};
```

packages/react/src/jsx-ast/utils/buildContent.mjs
```diff
@@ -317,6 +318,145 @@ export const processEntry = entry => {
...
+    }
+
+    current.children.shift();
+
+    if (activeOverloadGroup) {
+      processOverloadNode(current);
+      continue;
+    }
+
+    const last = finalChildren.pop();
```
ovflowd, line 438: nit: add inline comments on why we pop and shift.

ovflowd (approved):
SGTM! Now I swear final rounds of nits. After this we can merge!

avivkeller (approved):

moshams272:
> SGTM! Now I swear final rounds of nits. After this we can merge!

No, that's good for healthier & cleaner code.

ryzrr:
Amazing work @moshams272  btw,  since the overloads  doesnt have `<h3>`, will  they’ll break from search and just jump to the top? like maybe we should leave isOverload out of the Orama index in generate.mjs? That' keep the results clean too i think. WDYT?just a UX  thing nothing else ;)

packages/react/src/jsx-ast/utils/buildContent.mjs
moshams272: It works on the static AST arrays. That's where it is called: [diff](https://github.com/nodejs/doc-kit/pull/1047/changes#diff-bb448e05dc2e1a20cf7ec3b7853a3b5ee5450ed83890c278ff9406043bed4e46R479)

moshams272:
> Amazing work @moshams272 btw, since the overloads doesnt have `<h3>`, will they’ll break from search and just jump to the top? like maybe we should leave isOverload out of the Orama index in generate.mjs? That' keep the results clean too i think. WDYT?just a UX thing nothing else ;)

Nice catch 🤍 I was thinking of that, and here it's my report:

1. Solve this from root: we can just get the link of each tab and put it in the Orama, as when a user click on the search result that for the `Overload #3` as example, it'll redirect him to this tab dircetly, but now our component `CodeTabs` hasn't link at all. So we can open an upstream PR to solve this and then inject the links to the Orama here. It'll save the result, but it'll take a long time. (+1)

2. We can delete them(as u say), it's ez and doesn't take time, but we'll lose their data in search. (-0.5)

3. As we group them in one heading, why not group their links into this one heading also? it's ez, doesn't take time, and keeps their data. (+0.5)

WDUT, guys ?

ryzrr:
option 1 -> +1

ovflowd:
> option 1 -> +1

+1

moshams272:
I created an issue there: [issue at nodejs.org](https://github.com/nodejs/nodejs.org/issues/9140)

Waiting the approval...

PR #1049, opened 2026-08-21
fix(unist): convert commented markdown into HTML

avivkeller:
Resolves the issue shown below:
Screenshot 2026-08-21 at 9 26 26 AM

After:
Screenshot 2026-08-22 at 5 27 15 PM

### DIFF
--- a/packages/react/src/html/utils/config.mjs
+++ b/packages/react/src/html/utils/config.mjs
@@ -9,6 +9,7 @@ import {
 import { omitKeys } from '@doc-kit/core/utils/misc.mjs';
 import { LANGS } from '@node-core/rehype-shiki';
 
+import { toPlainText } from '../../jsx-ast/utils/buildContent.mjs';
 import { getSortedHeadNodes } from '../../jsx-ast/utils/getSortedHeadNodes.mjs';
 
 /**
@@ -58,7 +59,7 @@ export function buildDocumentationIndex(input) {
       api: entry.api,
       name: entry.heading.data.name,
       index: entry.stability.data.index,
-      description: getEntryDescription(entry),
+      description: toPlainText(getEntryDescription(entry)),
     }));
 }
 
--- a/packages/react/src/html/utils/config.mjs
+++ b/packages/react/src/html/utils/config.mjs
@@ -6,6 +6,7 @@ import {
   getEntryDescription,
   getVersionFromSemVer,
 } from '@doc-kit/core/utils/generators.mjs';
+import { parseInline, renderAsHTML } from '@doc-kit/core/utils/inline.mjs';
 import { omitKeys } from '@doc-kit/core/utils/misc.mjs';
 import { LANGS } from '@node-core/rehype-shiki';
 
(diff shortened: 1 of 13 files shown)


ovflowd (approved):

ovflowd:
@avivkeller is this ready for review?

avivkeller:
Not yet, but soon :-)

pnpm-workspace.yaml
```diff
@@ -3,3 +3,6 @@ packages:
 
 allowBuilds:
   unrs-resolver: false
+
+minimumReleaseAgeExclude:
+  - '@node-core/ui-components@1.7.6'
```
avivkeller, line 8: I'll remove this before landing, we just need to wait a few days...

packages/react/src/jsx-ast/utils/render.mjs
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Renders inline nodes as a JSX fragment
+ *
+ * @param {Array<import('mdast').PhrasingContent>} nodes - The nodes to render.
+ * @returns {import('estree-jsx').JSXFragment} The rendered nodes.
+ */
+export const renderAsJSX = nodes =>
+  remark().runSync(
+    createTree('root', [createJSXElement(null, { children: nodes })])
+  ).body[0].expression;
```
ovflowd, line 17: break down the [0] access and object .expression access

```tsx
const node = createTree('root', [createJSXElement(null, { children: nodes })]);

const [{ expression }] = remark().runSync(node);
```

packages/react/src/html/utils/config.mjs
```diff
@@ -58,7 +59,7 @@ export function buildDocumentationIndex(input) {
       api: entry.api,
       name: entry.heading.data.name,
       index: entry.stability.data.index,
-      description: getEntryDescription(entry),
+      description: renderAsHTML(parseInline(getEntryDescription(entry), true)),
```
ovflowd, line 62: What's the performance cost of this?

ovflowd (approved):
SGTM, left one comment :)

AugustinMauroy (approved):

packages/react/src/html/utils/config.mjs
avivkeller: Very little, it's a tiny tiny bit of markdown

packages/react/src/jsx-ast/utils/render.mjs
avivkeller: I copied this from the previous set up: https://github.com/nodejs/doc-kit/pull/1049/changes#diff-031004ae44cd1c0fe6d17dc040b1d74033ea782a68e7cccbe477e42135b97b26L73-L74

IMO changes can be in a follow-up

ovflowd: Does it matter that much? How hard is applying a nit? Will you really do it in a follow-up? I feel that this could easily be done here :melting_face:

PR #1050, opened 2026-08-22
fix(changesets): remove peer deps

avivkeller:
As titled, a fix for changesets

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -74,9 +74,9 @@
     "hast-util-to-string": "^3.0.1"
   },
   "peerDependencies": {
-    "@node-core/doc-kit": "workspace:*",
-    "@node-core/doc-kit-legacy": "workspace:*",
-    "@doc-kit/generator-react": "workspace:*"
+    "@node-core/doc-kit": "workspace:>=2.0.0",
+    "@node-core/doc-kit-legacy": "workspace:>=1.0.0",
+    "@doc-kit/generator-react": "workspace:>=0.1.0"
   },
   "peerDependenciesMeta": {
     "@node-core/doc-kit": {
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,8 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  }
 }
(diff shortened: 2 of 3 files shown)


packages/core/package.json
```diff
@@ -74,9 +74,9 @@
     "hast-util-to-string": "^3.0.1"
   },
   "peerDependencies": {
-    "@node-core/doc-kit": "workspace:*",
-    "@node-core/doc-kit-legacy": "workspace:*",
-    "@doc-kit/generator-react": "workspace:*"
```
avivkeller, line 88: The alternative here is to remove these from `peerDependencies`.

Anyone have a preference @nodejs/web-infra?

AugustinMauroy (approved):

ovflowd (approved):

packages/core/package.json
```diff
@@ -74,9 +74,9 @@
     "hast-util-to-string": "^3.0.1"
   },
   "peerDependencies": {
-    "@node-core/doc-kit": "workspace:*",
```
ovflowd, line 86: core itself doesn't depend on them right? So why are they on peer deps?

ovflowd: Just trying to understand what exactly is the rationale/logic behind this being here (not to mention it could created cyclical npm dependency?)

packages/core/package.json
bmuenzenmeyer: is it a fair characterization that they are here for our convenience, not true dependency? if so, they should be removed

avivkeller:
We need peer deps since without them, certain package managers (e.g. pnpm) will not link the dependencies to the core package.

ovflowd:
> We need peer deps since without them, certain package managers (e.g. pnpm) will not link the dependencies to the core package.

But shouldn't be the opposite? the other packages peerDepending on the core package?

avivkeller:
The other dependencies have a _required_ dependency of core, since they cannot function without them. Core has an optional peer dep, since it can function without them, but should they be installed, they need to be within this version range and a dep of core

ovflowd:
> The other dependencies have a _required_ dependency of core, since they cannot function without them. Core has an optional peer dep, since it can function without them, but should they be installed, they need to be within this version range and a dep of core

Right but I feel we are creating cyclically peerDeps core depends on their downstream [...]

ovflowd:
(maybe I am misunderstanding the implications of core having downstream deps depending on it as peerDeps)

avivkeller:
It _looks_ cyclic if you count peer edges, but it's an acyclic install graph, so this is an "intentional" semi-cycle (so-to-speak). core needs the peer declaration so pnpm resolves the optional packages into core's `node_modules` context.

PR #1057, opened 2026-08-27
feat(react): support chunking pages

avivkeller:
Chunks pages by their content, e.g. `fs/readFile`, etc.

`Set(['class', 'event', 'global', 'method']);` is the chunkable types, so a heading needs to satisfy it to be chunked.

In the process, improves performance by emitting a single JS entry for all pages, instead of one per page/chunk

**The preview is not chunked**


packages/react/src/html/bundlers/vite.mjs
```diff
@@ -202,7 +213,9 @@ export const createViteConfig = ({
     plugins: [
       createVirtualModulesPlugin(sources),
       ...(vite.plugins ?? []),
-      ...(!server && webConfig.minify ? [createHTMLFinalizerPlugin()] : []),
+      ...(!server && webConfig.minify
```
ovflowd, line 216: not a fan of this line

ovflowd (approved):
SGTM! One nit, could you rename the folder in source to `chunking` as a verb, since in the output `chunked` would be the chunked files?

packages/react/src/html/bundlers/vite.mjs
avivkeller: It makes sense to me, we are implementing a minifier

PR #1058, opened 2026-08-27
fix(core): respect log level in worker threads

### LINKED ISSUE #1032: doc-kit log level is not respected
```
 npx @doc-kit/cli generate -t web -o out -i /Users/user/Documents/node/doc/api/ --log-level fatal
[13:53:26.182] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1629:1629)
[13:53:26.208] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1636:1636)
[13:53:26.208] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1650:1650)
[13:53:26.208] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1682:1682)
[13:53:26.208] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1689:1689)
[13:53:26.208] WARN: Invalid type annotation: {HTTP/2 Headers Object} at /diagnostics_channel(1703:1703)
```

zeexzeex:
**What**

`--log-level` had no effect on warnings coming from worker threads.

```bash
npx @doc-kit/cli generate -t web -o out -i <node>/doc/api/ --log-level fatal
```

printed 81 `WARN` lines instead of none.

**Why**

The level is applied on the main thread only. Worker threads load their own module graph, so the default logger export creates a separate instance there that keeps the default `info` level. The warnings in the report come from `resolveTypes.mjs`, which runs inside the workers.

The same input makes it clear:

| Input | Workers used | default level | `--log-level fatal` |
| --- | --- | --- | --- |
| single file | no | 6 WARN | **0 WARN** |
| directory (`doc/api/`) | yes | 81 WARN | **81 WARN** |

Both inputs produce warnings, but only the single file respects the requested level. `parallel.mjs` skips the pool when `threads <= 1 || items.length <= 2`, so running the directory with `--threads 1` silences the warnings as well.

**How**

The level is passed through Piscina `workerData` and applied at the worker entry point. This adds a small `getLogLevel()` accessor, as there was no way to read the configured level back.

I also considered putting the level into `configuration`, which already reaches the workers, but `createTask` deliberately narrows what it sends:

```js
// Only pass the needed configuration to this generator
configuration: {
  [generatorName]: configuration[generatorName],
},
```

Sending the level that way would repeat it for every chunk, while `workerData` is sent once per worker.

**Verification**

| Case | Before | After |
| --- | --- | --- |
| `--log-level fatal` | 81 WARN | 0 WARN |
| default level | 81 WARN | 81 WARN |

The default level is unchanged, so nothing is silenced that should not be. Full suite passes (560 tests), lint and format checks are clean.

Fixes: https://github.com/nodejs/doc-kit/issues/1032

### DIFF
--- a/.changeset/worker-log-level.md
+++ b/.changeset/worker-log-level.md
@@ -1,0 +1,5 @@
+---
+'@doc-kit/core': patch
+---
+
+fix: respect `--log-level` in worker threads
--- a/packages/core/src/logger/__tests__/logger.test.mjs
+++ b/packages/core/src/logger/__tests__/logger.test.mjs
@@ -380,4 +380,35 @@ describe('createLogger', () => {
       strictEqual(transport.mock.callCount(), 1); // Debug should be filtered
     });
   });
+
+  describe('getLogLevel', () => {
+    it('should return the level the logger was created with', t => {
+      const logger = createLogger(t.mock.fn(), LogLevel.warn);
+
+      strictEqual(logger.getLogLevel(), LogLevel.warn);
+    });
+
+    it('should default to info when no level is given', t => {
+      const logger = createLogger(t.mock.fn());
+
+      strictEqual(logger.getLogLevel(), LogLevel.info);
+    });
+
+    it('should reflect a level set afterwards', t => {
+      const logger = createLogger(t.mock.fn(), LogLevel.info);
+
+      logger.setLogLevel('fatal');
+
+      strictEqual(logger.getLogLevel(), LogLevel.fatal);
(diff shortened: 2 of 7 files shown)


AugustinMauroy (approved):

bmuenzenmeyer (approved):
LGTM

PR #1077, opened 2026-09-02
style: update the width of history dropdown

### LINKED ISSUE #9145: Does the width of the changehistory item need to be adjusted?

Image

As shown in the image, the dropdown panel appears too wide when there's only one item. Should we change the `width` to `max-width`? However, this might cause the dropdown panel width to vary depending on the item.

btea:
## Description
/ 

## Related Issues

close nodejs/nodejs.org#9145
### Check List

### DIFF
--- a/packages/react/src/html/ui/index.css
+++ b/packages/react/src/html/ui/index.css
@@ -122,6 +122,12 @@ main {
       div[role='menu'] {
         left: 0;
 
+        /* Shrink the dropdown to fit its content instead of the fixed
+         * width set by @node-core/ui-components (nodejs/doc-kit#1051) */
+        width: fit-content;
+        min-width: 12rem;
+        max-width: 20rem;
+
         /* Prevent long labels from overflowing dropdown width */
         a[role='menuitem'] div {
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):

packages/react/src/html/ui/index.css
```diff
@@ -122,6 +122,12 @@ main {
       div[role='menu'] {
         left: 0;
 
+        /* Shrink the dropdown to fit its content instead of the fixed
+         * width set by @node-core/ui-components (nodejs/doc-kit#1051) */
+        width: fit-content;
+        min-width: 12rem;
+        max-width: 20rem;
```
avivkeller, line 129: Perhaps we update this in ui-components?

btea: I'm unsure whether to make uniform changes within the component or to implement targeted changes within projects that use the component.

ovflowd: IMO should be on ui-components itself too!

PR #1078, opened 2026-09-02
fix: normalize paths in Windows tests

divyanshisingh987456321:
## Description

Normalize file paths to use forward slashes in output comparisons, ensuring tests behave consistently on Windows.

This updates the affected generators and comparators to avoid Windows-specific backslash paths in test output.

## Validation

- [x] Ran `node --run test` — all 598 tests passed.
- [x] Ran `node --run format:check` — passed.
- [x] Ran `node --run lint` — passed.

## Related Issues

N/A

### Check List

- [x] I have check code formatting with `node --run format:check` & `node --run lint`.

### DIFF
--- a/packages/core/src/generators/ast/generate.mjs
+++ b/packages/core/src/generators/ast/generate.mjs
@@ -1,7 +1,7 @@
 'use strict';
 
 import { readFile } from 'node:fs/promises';
-import { relative, sep } from 'node:path/posix';
+import { relative, sep } from 'node:path';
 
 import globParent from 'glob-parent';
 import { globSync } from 'tinyglobby';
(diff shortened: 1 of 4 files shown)


packages/core/src/generators/ast/generate.mjs
```diff
@@ -65,7 +65,7 @@ export async function processChunk(inputSlice, itemIndices) {
     );
 
     // The path is the relative path minus the extension
-    const relativePath = sep + withExt(relative(parent, path));
+    const relativePath = `/${withExt(relative(parent, path)).replaceAll(sep, '/')}`;
```
avivkeller, line 68: Isn't this the exact same?

ovflowd: It sounds like it is doing the opposite, converting any sep into `/` but that feels ... wrong? In Windows the `\` sep should be used and not the `/` one :sweat_smile:

bmuenzenmeyer: this feels backwards - sep is the right way to do this AFAIK

PR #1079, opened 2026-09-02
feat(json): add JSON generators

### LINKED ISSUE #214: New JSON generator schema

avivkeller:
This PR is meant to supersede https://github.com/nodejs/doc-kit/pull/287 and fix https://github.com/nodejs/doc-kit/issues/214.

The majority of the added lines are test fixtures and schemas, so do not be alarmed by the size of this PR.

This new generator writes one JSON document per source file. Each document
is a tree of the file's headings, in document order, with the metadata,
signature or type, Markdown body, and code examples of every one of them.

The output is described by a JSON schema, published at the URL every document
carries in `$schema`, and shipped with the package as
`@doc-kit/core/generators/json/schema.json`.

An output looks roughly like:
```json
{
  "$schema": "https://doc-kit.nodejs.org/schemas/api-doc/1.0.0.json",
  "id": "fs",
  "path": "/fs",
  "type": "module",
  "module": "fs",
  "title": "File system",
  "introducedIn": "v0.10.0",
  "sourceLink": {
    "path": "lib/fs.js",
    "url": "https://github.com/nodejs/node/blob/HEAD/lib/fs.js"
  },
  "stability": { "index": "2","description": "Stable" },
  "added": [],
  "deprecated": [],
  "removed": [],
  "napiVersion": [],
  "changes": [],
  "description": "The `node:fs` module enables interacting with the file system in a\nway modeled on standard POSIX functions.\n\n…",
  "summary": "The `node:fs` module enables interacting with the file system in a way modeled on standard POSIX functions.",
  "examples": [],
  "children": []
}
```


packages/core/src/generators/json-all/constants.mjs
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+// Where a version of the bundle's schema is published.
+export const SCHEMA_URL =
+  'https://doc-kit.nodejs.org/schemas/api-doc-all/{schemaVersion}.json';
```
AugustinMauroy, line 5: IMO we should generate that and give it with the generated artifact so a built can be served independently

avivkeller: They are welcome to export the artifact from the source code and host it independently, we _also_, by default, host our schema

packages/core/src/generators/json/__tests__/generate.test.mjs
AugustinMauroy, line 1: can we have an other fixtures thats look like deprecation.

I use this script that read json so I would like to see output of such as file https://github.com/brunocroh/node-deprecations-pretty

packages/core/src/generators/json/README.md
AugustinMauroy, line 1: we should include the ts file link

packages/core/src/generators/json/__tests__/schema.test.mjs
```diff
@@ -0,0 +1,36 @@
+import assert from 'node:assert/strict';
+import { readFile } from 'node:fs/promises';
+import { describe, it } from 'node:test';
+
+import Ajv from 'ajv';
```
ovflowd, line 5: I assume this is a test-only (dev) dependency, right?

packages/core/src/generators/json/generated/schema.d.ts
```diff
@@ -0,0 +1,351 @@
...
+ */
+export type Version = string;
+/**
+ * A heading below the document root.
+ */
+export type Node = SectionNode | ClassNode | ConstructorNode | MethodNode | StaticMethodNode | PropertyNode | EventNode;
+/**
+ * A heading that documents no API entry: prose, a deprecation, a command-line option.
+ */
+export type SectionNode = Entry &
```
ovflowd, line 55: Can you use interface in these scenarios (SectionNode, ClassNode, ConstructionNode...), so we do use `extends`?

ovflowd: (I know this is generated, but wondering why a generated schema lives on git hmmm)

packages/core/src/generators/json/utils/document.mjs
```diff
@@ -0,0 +1,58 @@
...
+  annotateOverloads(entries);
+
+  const type = DOCUMENT_TYPES.has(head.type) ? head.type : 'module';
+  const scope = type === 'global' ? 'global' : 'module';
+
+  // The title entry is the root; any other root (a second title, an entry
+  // with no shallower heading before it) is treated as a child of it
+  const roots = buildHierarchy(entries);
+  const root = roots.find(node => node.entry === head) ?? roots[0];
+  const children = [
```
ovflowd, line 28: nit: assign const for each one of these things instead of several assignment + spreads

packages/core/src/generators/json/utils/document.mjs
```diff
@@ -0,0 +1,58 @@
...
+      typeof head.name === 'string'
+        ? head.name
+        : type === 'module'
+          ? head.api
+          : null,
+    title,
+    introducedIn:
+      head.introduced_in == null ? null : String(head.introduced_in),
+    sourceLink: head.source_link
+      ? {
```
ovflowd, line 50: nit: fn just for generating this sourceLink property

packages/core/src/generators/json/utils/entry.mjs
```diff
@@ -0,0 +1,87 @@
...
+    !items.some(item => extractListItem(item).prefix === 'Extends')
+  ) {
+    return { body, items: [] };
+  }
+
+  return {
+    body: body.toSpliced(
+      index,
+      1,
+      ...(rest.length ? [{ ...list, children: rest }] : [])
```
ovflowd, line 63: nit: avoid spread + assignment

packages/core/src/generators/json/utils/entry.mjs
```diff
@@ -0,0 +1,87 @@
...
+  if (index === -1) {
+    return { body, items: [] };
+  }
+
+  const list = body[index];
+  const { typed: items, rest } = splitTypedItems(list);
+
+  if (
+    kind === 'class' &&
+    !items.some(item => extractListItem(item).prefix === 'Extends')
```
ovflowd, line 54: How many items can be in this Array? `.some` is not that performant...

packages/core/src/generators/json/utils/node.mjs
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * The properties a node has on top of its entry, by kind.
+ *
+ * @param {string} kind
+ * @param {import('../../metadata/types').MetadataEntry} entry
+ * @param {Array<import('mdast').ListItem>} items The entry's typed list items
+ * @returns {{ properties: object, description?: string }}
+ */
+const kindProperties = (kind, entry, items) => {
```
ovflowd, line 74: I wonder if instead of a switch we could use a factory, might be overkill, but better specified and controled/standardized.

packages/core/src/generators/json/utils/node.mjs
```diff
@@ -0,0 +1,153 @@
...
+    title,
+    scope: entryScope,
+    overloadOf: entry.heading.data.overloadOf ?? null,
+    stability,
+    added,
+    deprecated,
+    removed,
+    napiVersion,
+    changes,
+    ...properties,
```
ovflowd, line 146: Why the spread happens before a speciifc item? It'd be good to know what could be accidentally overridden here and that it is fragile that it needs to be in a speciific position.

packages/core/src/generators/json/utils/signature.mjs
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @param {Array<import('mdast').ListItem>} items The typed list's items
+ * @param {boolean} [mdx] Whether the items come from an MDX document
+ * @returns {{ type: import('../types').Type | null, default: string | null, description: string }}
+ */
+export const buildPropertyType = (items, mdx) => {
+  const [item] = items.map(extractListItem);
+
+  return item
+    ? {
```
ovflowd, line 169: I know it might not sound practical, but could you remove the ternary and make it like below (easier to read) 

```ts
if (item) {
 return { ... }
}

return { type: null, ... }
```

packages/core/src/generators/json/utils/types.mjs
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+/**
+ * Builds a type from a resolved `typeAnnotation` node.
+ *
+ * @param {import('mdast').Node | undefined} node
+ * @returns {import('../types').Type | null}
+ */
+export const toType = node =>
+  node
```
ovflowd, line 10: same here... this is the sort of situation I'd rather NOT use a ternary

packages/core/src/generators/json/constants.mjs
```diff
@@ -0,0 +1,57 @@
...
+]);
+
+// The document-level `type` classifications
+export const DOCUMENT_TYPES = new Set(['module', 'misc', 'global']);
+
+// The `displayName="..."` attribute of a fenced code block's info string
+export const DISPLAY_NAME = /displayName="([^"]*)"/;
+
+// A rest parameter's marker
+export const REST_MARKER = /^\.\.\./;
```
ovflowd, line 48: Are these RegEx's available elsewhere? Are these duplication of already existing RegEx's?

packages/core/src/generators/json/constants.mjs
```diff
@@ -0,0 +1,57 @@
...
+  classMethod: 'staticMethod',
+  property: 'property',
+  event: 'event',
+};
+
+// The kind of a heading that documents no API entry
+export const SECTION_KIND = 'section';
+
+// The kinds whose leading typed list is lifted out of the body as data
+export const KINDS_WITH_TYPED_LIST = new Set([
```
ovflowd, line 25: Also wondering if this is already defined elsewhere

packages/core/src/utils/stability.mjs
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {import('../generators/metadata/types').StabilityNode} node
+ * @returns {import('mdast').Blockquote | null} The copy, or `null` when nothing follows the prefix
+ */
+export const removeStabilityPrefix = node => {
+  const text = toString(node.children[0]);
+  const match = QUERIES.stabilityIndex.exec(text);
+  const start = match ? text.length - match[2].length : 0;
+
+  return slice(node, start, undefined, {
```
ovflowd, line 21: nit: instead `.node` -> `const { node } = `

scripts/generate-json-types.mjs
```diff
@@ -0,0 +1,51 @@
...
+export const compileSchemaTypes = async () => {
+  const schema = JSON.parse(await readFile(SCHEMA_PATH, 'utf-8'));
+
+  // The repository's formatting, minus the per-path overrides
+  const style = JSON.parse(await readFile(PRETTIER_CONFIG, 'utf-8'));
+  delete style.overrides;
+
+  return compile(schema, 'Document', {
+    bannerComment:
+      '/* eslint-disable */\n' +
```
ovflowd, line 36: Should this be a constant?

package.json
```diff
@@ -44,6 +44,7 @@
     "eslint-plugin-react-x": "5.18.1",
     "globals": "~17.7.0",
     "husky": "9.1.7",
+    "json-schema-to-typescript": "^16.0.0",
```
ovflowd, line 47: yet another package... At least it is a dev one, so Node.js consuming doc-kit won't use it right?

ovflowd (commented):
Looking really good. Left a few comments, I left out more specific nits, to let this move forward. If you could address my current comments I believe I'm good with a ✔️

ovflowd:
Can you also rebase, @avivkeller?

packages/core/src/generators/json/__tests__/schema.test.mjs
avivkeller: Yes

packages/core/src/generators/json/generated/schema.d.ts
avivkeller: I don't think this can be changed manually, it's auto-generated from `json-schema-to-typescript`. As for why it's committed, it's to avoid making `json-schema-to-typescript` a production dependency.

packages/core/src/generators/json/utils/entry.mjs
avivkeller: > `.some` is not that performant...

No more than a few (less than 5 most likely)

packages/core/src/generators/json/utils/node.mjs
avivkeller: > It'd be good to know what could be accidentally overridden here and that it is fragile that it needs to be in a speciific position.

We control the value of `properties`, so nothing can be overridden. The position of the spread in the object is irrelevant, I can move it if it worries you.

packages/core/src/generators/json/utils/signature.mjs
avivkeller: I think this is more of a personal preference, i.e. I prefer ternaries. I'm going to leave this as is, and if we find it's a bigger problem, we can add a lint rule

packages/core/src/generators/json/constants.mjs
avivkeller: It's not

package.json
avivkeller: Correct

packages/core/src/generators/json/utils/entry.mjs
avivkeller: It's a stylistic thing, I think it's more readable to understand what's spreading and what's not

packages/core/src/generators/json/utils/document.mjs
avivkeller: I think it's fine like this, no need to instantiate variables for no reason

packages/core/src/generators/json/utils/node.mjs
ovflowd: Yeah, let's put it as the last item, just to ensure it is not conflicting with anything, and if it ever does, it becomes easy to know

ovflowd (approved):
One last comment to be addressed, feel free to merge after that. SGTM!

packages/core/src/generators/json-all/schema.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://doc-kit.nodejs.org/schemas/api-doc-all/1.0.0.json",
```
bmuenzenmeyer: question: should this schema lockstep match https://doc-kit.nodejs.org/specification so there is never a question of compatibility mapping?

packages/core/src/generators/json-all/schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "title": "Bundle",
+  "description": "Every document of a documentation set, as emitted by the doc-kit `json-all` generator, in index order.",
```
bmuenzenmeyer: nit: other ordering comments mention index order _and then_ id order.

packages/core/src/generators/json-all/types.d.ts
```diff
@@ -0,0 +1,20 @@
+import type { Document } from '../json/generated/schema';
+
+/**
+ * Every document of a documentation set, in index order.
```
bmuenzenmeyer: nit: no mention of the other order. ignore if not relevant. just seeing different fidelities of info throughout

packages/core/src/generators/json-simple/generate.mjs
```diff
@@ -23,10 +23,7 @@ export async function generate(input) {
   if (config.output) {
     // Writes all the API docs stringified content into one file
     // Note: The full JSON generator in the future will create one JSON file per top-level API doc file
```
bmuenzenmeyer: this is no longer a future statement

packages/core/src/generators/json/generated/schema.d.ts
bmuenzenmeyer, line 1: should this be checked in if generated?

packages/core/src/generators/json/constants.mjs
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+// See the `version` in schema.json. These values should be in sync.
+export const SCHEMA_VERSION = '1.0.0';
```
bmuenzenmeyer: another place to consider https://doc-kit.nodejs.org/specification

packages/core/src/generators/json/constants.mjs
bmuenzenmeyer: this seems like a valid question considering some of our marketing for doc kit has been "less regex, more AST"

packages/core/src/generators/json/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+The output is described by a JSON schema, published at the URL every document
+carries in `$schema`, and shipped with the package as
+`@doc-kit/core/generators/json/schema.json`.
+
+```sh
+npx @doc-kit/cli generate -t json -i "doc/api/*.md" -o out
+```
+
+Files keep the input's directory layout: `doc/api/fs.md` becomes `out/fs.json`.
```
bmuenzenmeyer: This statement seems to contradict its example

README.md
```diff
@@ -69,10 +69,11 @@ Options:
...
-                               (json-simple, legacy-html, legacy-html-all,
-                               man-page, legacy-json, legacy-json-all,
-                               addon-verify, api-links, orama-db, llms-txt,
-                               sitemap, html) or an import specifier for a
+                               (json, json-all, json-simple, legacy-html,
+                               legacy-html-all, man-page, legacy-json,
+                               legacy-json-all, addon-verify, api-links,
+                               orama-db, llms-txt, sitemap, html,
+                               section-pages) or an import specifier for a
                                custom generator
```
bmuenzenmeyer, line 77: nit: this list is not in a sensical order to me

packages/core/src/generators/json/constants.mjs
avivkeller: It is "less regex", except for simple ones, like `...`

README.md
avivkeller: It's vertabrim from the help menu

bmuenzenmeyer (approved):
I only wonder if we should add something to the specification or CONTRIBUTING.md to now mention that the schema version spans multiple location

PR #1082, opened 2026-09-04
feat: add `preview:docs` & `preview:node` scripts

btea:
## Description
Add a unified startup preview script to facilitate quick previewing and verification of changes on the page after modifying the code logic.

### Check List

### DIFF
--- a/scripts/preview.mjs
+++ b/scripts/preview.mjs
@@ -1,0 +1,140 @@
+#!/usr/bin/env node
+
+// Builds one of the two doc sites in this repository, then serves it locally:
+//
+//   node scripts/preview.mjs node  -> ./out       served on http://localhost:3000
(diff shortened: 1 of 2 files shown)


scripts/preview.mjs
ovflowd, line 1: Out of curiosity, is this meant to facilitate local development? I think in the past we decided against such scripts. I do believe we could make _instructions_ easier tho.

btea: Yes, modifying the code locally and then validating it on the page requires building the latest code and starting a service. This process is quite cumbersome and involves many parameters, so I'm thinking of adding a script to simplify the process.

bmuenzenmeyer (approved):
I've come around to this after observing a mentee interact with this repository. There is a lot of intimidating work to get all this to work together, and for the unfamiliar, it's a lot.

AugustinMauroy (approved):

PR #1090, opened 2026-09-06
perf: improve perf with more delegation

### LINKED ISSUE #1008: Make `all.html` a concatenation of existing pages
Currently, generating the all page requires data from, well, all pages, making it slow and high memory (since all the pages need to remain in the memory). If we could adapt our pipeline somehow to make this page able to use the existing pages and simply append them together, it would make this process much faster and much less memory intensive.

Keep in mind this is a difficult task, despite the seemingly simplicity of this issue. Since we output pages complete with sidebars, navbars, etc, one can't simple do `html + html + html`.

avivkeller:
Fixes #1008

- The `html` generator builds the component library once and the client assets
  once, then compiles, renders, minifies and writes each page on its own in the
  worker pool, so memory scales with the largest page rather than with the
  site. On the Node.js API docs with section pages, peak memory drops from
  about 8 GB to 3.4 GB and the build runs in half the time.
- `all.html` is assembled from the module pages' compiled content, in sidebar
  order, instead of being built again from every module. The option moved from
  `jsx-ast.generateAllPage` to `html.generateAllPage`, and the page shows no
  reading time.
- `jsx-ast` emits `{ data, headings, readingTime, content }` per page,
  `content` being the page body as one JSX fragment; `html` wraps it in
  `<Layout>`.
- The `WebBundler` contract is now `buildServer` / `compile` / `buildClient`
  (see the `html` README). Vite plugins no longer see the HTML pages; customize
  them through the template, whose `${entrypoint}` is replaced by `${assets}`.
  Synthetic pages load assets from the site root, so `404.html` works at nested
  paths.
- `@doc-kit/core`: `getRemarkRehypeWithShiki` moved to
  `@doc-kit/core/utils/remark-shiki.mjs` and `typeAnnotationToHighlightedHast`
  to `@doc-kit/core/utils/type-annotations/highlighted.mjs`, so the `ast` and
  `metadata` stages no longer load Shiki. The default `threads` is capped at 4.
  Function-valued generator configuration (such as a custom `bundler`) is
  dropped before it is sent to workers instead of failing the structured clone.


packages/core/src/utils/configuration/index.mjs
```diff
@@ -20,6 +20,9 @@ import { deepMerge } from '#utils/misc.mjs';
 
 const configExplorer = cosmiconfig('doc-kit');
 
+// The default `threads` ceiling; `--threads` raises it explicitly.
+const MAX_THREADS = 4;
```
ovflowd: nit: That should be in a constant file

packages/react/src/jsx-ast/generate.mjs
```diff
@@ -5,34 +5,27 @@ import { jsx, toJs } from 'estree-util-to-js';
...
- * `all` page is never built on the main thread.
- *
- * @param {Array<import('@doc-kit/core/generators/metadata/types').MetadataEntry>} input
+ * Builds the `{ head, entries }` page descriptors for the configured synthetic
+ * pages. `all.html` is not one of them: it is the module pages concatenated,
+ * so the `html` generator assembles it from their content instead of building
+ * every module a second time here.
  */
-const buildSyntheticDescriptors = input => {
+const buildSyntheticDescriptors = () => {
```
ovflowd: This name.... doesn't mean much, now this could just be buildNotFoundPage imo

packages/react/src/html/bundlers/vite.mjs
ovflowd, line 1: These changes seem pretty good, any benchmarks/numbers on the gains here?

packages/react/src/html/bundlers/vite.mjs
```diff
@@ -337,109 +274,141 @@ export const createViteConfig = ({
...
+      outDir,
+      config,
+      vite,
+    })
+  );
 
-  // Vite writes the compiled SSR renderers here so Node can import and execute
-  // them without mixing intermediate modules into the final site. The directory
-  // is removed after rendering
-  const temporaryDirectory = await createTemporaryDirectory();
```
ovflowd, line 366: No need more for a temporary dir _or_ is that now managed somewhere else?

packages/react/src/html/bundlers/vite.mjs
```diff
@@ -337,109 +274,141 @@ export const createViteConfig = ({
...
-  const root = resolve(vite.root ?? process.cwd());
-  const input = [];
-
-  for (const [fileName, html] of pages) {
-    const id = resolve(root, fileName);
-    input.push(id);
-    sources.set(id, html);
-  }
+  const sources = new Map([
+    [
```
ovflowd, line 368: what is this import module? Is this to allow HMR?

ovflowd: Should this _always_ be bundled with the server?

packages/react/src/html/bundlers/vite.mjs
```diff
@@ -337,109 +274,141 @@ export const createViteConfig = ({
...
+      input: { [CLIENT_NAME]: CLIENT_ENTRY_ID },
       server: false,
-      minifyPages,
+      outDir,
       config,
       vite,
     })
   );
+
+  const requested = vite.build?.manifest;
```
ovflowd: ooc, why the name of this variable is requested, is there a reason? just trying to understand if the name is intentional :P

packages/react/src/html/bundlers/vite.mjs
```diff
@@ -337,109 +274,141 @@ export const createViteConfig = ({
...
+    await rm(manifestPath);
+    await rmdir(dirname(manifestPath)).catch(() => {});
+  }
+
+  const chunk = Object.values(manifest).find(item => item.isEntry);
+
+  return {
+    scripts: [chunk.file],
+    preloads: collectImports(manifest, chunk),
+    // With CSS code splitting off, the one stylesheet is its own manifest
```
ovflowd: not sure I understand, could you rephrase for clarity?

packages/react/src/html/utils/all.mjs
```diff
@@ -0,0 +1,42 @@
...
+ * Modules are ordered as the sidebar lists them.
+ *
+ * @param {Array<import('../../jsx-ast/types').PageCode>} pages - Every generated page
+ * @returns {import('../types').ComposedPage}
+ */
+export const buildAllPage = pages => {
+  const byApi = new Map(pages.map(page => [page.data.api, page]));
+
+  const parts = getSortedHeadNodes(
+    pages
```
ovflowd: Could you assign this to its own const, making this line simpler?

packages/react/src/html/utils/all.mjs
```diff
@@ -0,0 +1,42 @@
...
+      .map(({ data }) => data)
+      .filter(data => !data.synthetic && !data.chunk && data.api !== 'index')
+  ).map(({ api }) => byApi.get(api));
+
+  const minutes = parts.reduce(
```
ovflowd: I'd argue we shouldn't have a reading minutes on our API docs, only on Learn/Blog content. Nor should we have it on all.html

packages/react/src/html/utils/generate.mjs
```diff
@@ -59,39 +92,66 @@ export default () => {
...
-        false
-      ),
-
-      // Export a renderer that the server bundler can execute.
-      `export default () => renderToStringAsync(${componentCode});`,
+      `import { ${[...RUNTIME_IMPORTS, ...used].join(', ')} } from ${quote(libraryURL)};`,
+      ...imports,
+      `export const headings = ${JSON.stringify(headings)};`,
+      `export const content = () => ${content};`,
+      `export default () => renderToStringAsync(<${JSX_IMPORTS.Layout.name} metadata={${JSON.stringify(metadata)}} headings={headings} readingTime={${JSON.stringify(readingTime?.text)}}>{content()}</${JSX_IMPORTS.Layout.name}>);`,
```
ovflowd: Agree, I'd aruge sanitization would be good here

packages/react/src/html/utils/processing.mjs
```diff
@@ -116,113 +114,75 @@ export const buildHead = ({ meta = [], links = [], html = [] }) =>
...
+ * page's root: the entry script, the chunks it statically imports (preloaded,
+ * as the bundler would), and the stylesheets.
  *
- * @returns {{ add: (item: { data: import('@doc-kit/core/generators/metadata/types').MetadataEntry, code: string }) => void, serverCodeMap: Map<string, string>, clientProgram: string }}
+ * @param {import('../types').ClientAssets} assets - Output-relative asset paths
+ * @param {string} root - The page's root (see {@link resolvePageRoot})
+ * @returns {string}
+ */
+export const buildAssetTags = ({ scripts, preloads, stylesheets }, root) =>
+  [
```
ovflowd, line 127: Could this not be a multi level spread? Also is string manipulation the best way of doing this?

ovflowd (approved):
I guess the part that makes this hard to review is since this has an agglomeration of a lot of different changes in several files, it makes it _hard_ to know if for example, on vite.mjs are all changes there related to one thing? multiple things? It makes it harder to actually use Git as VCS and properly isolate changes.

I really appreciate if we could separate each change, PR stacks seems a good way to be able to get this done all at once reasonably, wdyt?

PR #1093, opened 2026-09-08
chore: enable `minimumReleaseAgeExcludePrune`

btea:
## Description
https://pnpm.io/11.x/settings/dependency-resolution#minimumreleaseageexcludeprune

### Check List

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -5,5 +5,3 @@ allowBuilds:
   unrs-resolver: false
 
 minimumReleaseAgeExcludePrune: true
-minimumReleaseAgeExclude:
-  - '@node-core/ui-components@1.7.6'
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -4,5 +4,6 @@ packages:
 allowBuilds:
   unrs-resolver: false
 
+minimumReleaseAgeExcludePrune: true
 minimumReleaseAgeExclude:
   - '@node-core/ui-components@1.7.6'
```
avivkeller: Can you also prune these

btea: Done. At this point, manual removal is the only option.

avivkeller (approved):

bmuenzenmeyer (approved):

MattIPv4 (approved):
Surprised this isn't a default.

Also, we don't have a workflow here to update pnpm automatically like our other repos?

PR #1103, opened 2026-09-15
fix(react): resolve theme toggle hydration mismatch

puneetnith28:
## Description

Fixes a hydration mismatch where the `ThemeToggle` icon resets to the `system` icon after page reload when a user preference (`dark` / `light`) is stored.

## Changes

- Initialized `pref` state to `'system'` by default in `useTheme.mjs` to ensure server HTML and client initial render match.
- Synchronized `pref` with `localStorage` inside a `useEffect` hook on client mount.
- Added `key={themePreference}` to `<ThemeToggle>` to ensure clean remounting when the theme preference is applied.

## References

- **Related Issue:** [webpack/webpack-doc-kit#269](https://github.com/webpack/webpack-doc-kit/issues/269)
- **Related PR:** [webpack/webpack-doc-kit#270](https://github.com/webpack/webpack-doc-kit/pull/270)


puneetnith28:
Please review this PR and let me know if any further changes are needed.
Thanks!!

packages/react/src/html/ui/hooks/useTheme.mjs
```diff
@@ -29,10 +29,14 @@ const applySystemTheme = () => applyTheme('system');
...
-    server ? 'system' : (localStorage.getItem('theme') ?? 'system')
-  );
+  const [pref, setPref] = useState('system');
+
+  useEffect(() => {
+    const stored = localStorage.getItem('theme');
+    if (stored) {
+      setPref(stored);
+    }
+  }, []);
```
TusharThakur04, line 39:
```suggestion
useEffect(() => {
  const stored = localStorage.getItem('theme');

  if (stored === 'light' || stored === 'dark') {
    setPref(stored);
  }
}, []);
```

TusharThakur04: we dont need to reapply `system`

avivkeller: wouldn’t `stored !== pref` be a little more robust?

TusharThakur04: if somehow `stored` is returned as `null`, its better to check explicitly. WDYT?

puneetnith28: Updated to explicitly check for 'light' or 'dark' to safely handle null and avoid re-applying system. 
Thanks for suggestion !!

packages/react/src/html/ui/components/ThemeToggle.jsx
```diff
@@ -10,7 +10,11 @@ const Toggle = () => {
   const [themePreference, setThemePreference] = useTheme();
 
   return (
-    <ThemeToggle onChange={setThemePreference} currentTheme={themePreference} />
+    <ThemeToggle
+      onChange={setThemePreference}
+      currentTheme={themePreference}
+      key={themePreference}
```
ryzrr: why  we  need key here?  without this , it seems to fix the icon just the same

TusharThakur04 (approved):

packages/react/src/html/ui/components/ThemeToggle.jsx
puneetnith28: I added key thinking a remount might be needed, but currentTheme prop updates already handle the re-render properly

ryzrr:
LGTM !

AugustinMauroy (approved):

avivkeller (approved):

PR #1104, opened 2026-09-16
fix(react): avoid empty label segments and duplicate keys

btea:
## Description
Avoid rendering empty content tags.

### Check List


packages/react/src/html/ui/utils/renderLabel.jsx
```diff
@@ -16,11 +16,22 @@ export const renderLabel = label => {
...
-      <code key={`code:${segment}`}>{segment}</code>
+  const segmentCounts = new Map();
+
+  return segments.map((segment, index) => {
+    if (!segment) {
+      return null;
+    }
+
+    const type = index % 2 ? 'code' : 'text';
+    const count = (segmentCounts.get(`${type}:${segment}`) ?? 0) + 1;
```
btea: If the strings being split are duplicated, it could result in duplicate keys, so try to avoid that (even though the chances are slim).

AugustinMauroy (approved):

packages/react/src/html/ui/utils/renderLabel.jsx
```diff
@@ -16,11 +16,22 @@ export const renderLabel = label => {
...
+
+    segmentCounts.set(`${type}:${segment}`, count);
+
+    return index % 2 ? (
+      <code key={`${type}:${segment}:${count}`}>{segment}</code>
     ) : (
-      <span key={`text:${segment}`}>{segment}</span>
-    )
-  );
+      <span key={`${type}:${segment}:${count}`}>{segment}</span>
```
avivkeller: The chances of this happening are highly unlikely, it makes more sense to tie `key` to `index`

btea: Done.

avivkeller (approved):

